PR #4, opened 2026-06-09
feat: Step 0 core layer — types, seeds, chain wrappers

mega123-art:
## Summary

Foundation for T2 (on-chain skills): core layer (Step 0) + skill NFT (Step 3) + notes + search (Step 4) + reputation (Step 5) + validation adapter (Step 6) + workflow NFTs (Step 7) + MCP Tools (Step 8).

### Step 0 — Core Layer

- **core/types.ts** — domain types (Session, Skill, Workflow, Note, Row)
- **core/seed.ts** — table hints + constants (AGENTNET_ROOT_ID, mysessionsHint, notesSkillHint, notesAgentHint, AUDIT_HINT)
- **core/chain.ts** — solana-sdk wrappers (init, ensureDbRoot, createTable, writeRow, readRows, codeIn, readCodeIn, tableExists, signerAddress)

### Step 3 — Skill NFT (Token-2022)

- **nft/token2022.ts** — createSkillMint (NonTransferable mint), mintSkillToken (soulbound minting)
- **nft/skill.ts** — publishSkill (code-in text → mint with uri=txid), buySkill (atomic: transfer payment + mint)

Semi-fungible Token-2022: one mint per skill, supply = popularity counter. NFT layer (Token-2022) and content layer (IQLabs code-in) decoupled via uri field. buySkill unifies star + payment + equip in one atomic tx.

### Step 4 — Notes + Search

- **notes/notes.ts** — postNote (gated by token balance ≥1), readNotes, deleteNote (stub)
- **notes/balance.ts** — getBalance helper (check wallet owns skill token)
- **search/search.ts** — searchSkills with filters (keyword, category, hashtags, type) and sort (supply desc, name asc, recent desc)

On-chain comments keyed by skill mint. postNote verifies author owns ≥1 skill token before allowing write. Search reads from AUDIT_HINT, filters client-side by keyword (name/description), category (exact), hashtags (any match), and type (skill vs workflow). Sorts by supply (popularity), name, or timestamp.

### Step 5 — Reputation + Leaderboard

- **reputation/reputation.ts** — getReputation (aggregates skills published, total supply, notes received → score), updateReputation (writes snapshot on-chain), getLeaderboard (sorted by score)

Score formula: `(totalSupply × 3) + (skillsPublished × 10) + notesReceived`. Leaderboard avoids N+1 by omitting notes count for ranking reads.

### Step 6 — Skill Validation Adapter

- **nft/validation/types.ts** — `ValidationAdapter`, `ValidationResult`, `Issue`, `ValidationError`
- **nft/validation/parse.ts** — safe YAML frontmatter parser (no eval, CWE-150 sanitizer), SPDX + URL helpers, extracts `type` and `requiredSkills`
- **nft/validation/adapters/compat.ts** — minimal skills.sh compat (name + description required)
- **nft/validation/adapters/strict.ts** — PR #509 quality rules (kebab-case name, description length, SPDX license, URL repository)
- **nft/validation/adapters/onchain.ts** — on-chain constraints (≤700B size limit, category, hashtag hygiene); exported as `defaultValidator`
- **nft/validation/adapters/security.ts** — injectable LLM maliciousness adapter (reviewFn caller-provided, fail-open on outage)
- **nft/validation/compose.ts** — chain adapters into a pipeline (fail-fast or collect-all mode)
- **nft/skill.ts** — `publishSkill` now runs validation gate as step 0 before any chain call; accepts optional custom `validator`

Pluggable adapter stack: compat → strict → onchain → security-llm. Default gate (OnchainAdapter) blocks invalid skills before they touch the chain. Security adapter is model-agnostic via injectable `reviewFn`.

### Step 7 — Workflow NFTs

- **nft/workflow.ts** — `publishWorkflow` (codes-in recipe, mints token), `unlockWorkflow` (verifies wallet holds all constituent skills before minting)
- **nft/validation/adapters/workflow.ts** — `WorkflowAdapter` enforces `type: workflow` and verifies `requiredSkills` is a non-empty array of valid base58 mints.

Workflows use the exact same Token-2022 soulbound architecture as skills. The primary distinction is the **Prerequisite Gate**: `unlockWorkflow` loops through all required skill mints and asserts `getBalance >= 1n` for each before constructing the mint transaction. 

### Step 8 — MCP Tools for Autonomous Buy

- **mcp/server.ts** — Configures the standard `@modelcontextprotocol/sdk` `Server` class, exposing `search_skills` and `buy_skill` as tools to any connected AI agent. It allows agents to perform autonomous actions inside the ecosystem.
- **mcp/skill-shopping.md** — Natural-language workflow instructions that act as a template. Instructs agents on how to identify missing skills mid-task and use the MCP tools to query the AgentNet market and acquire them.

## Status

All steps compile clean. 94 unit tests passing across 17 suites. Ready for review.


mega123-art:
## Design note — two open decisions: DAS enumeration & on-chain note gating

Flagging two things that are **open design decisions**, not coding gaps, so they don't get silently wired the wrong way. Both trace back to one root: skills are **standalone Token-2022 mints** with no grouping primitive, and the deployed IQ contract's gate is **legacy-SPL only**.

### 1. DAS enumeration ("no registry table", skill-nft-structure.md §2)

The plan wants the NFT collection to *be* the skill list, enumerated via DAS. Verified against the SDK:

- The IQ solana-sdk uses **no DAS getAssets** methods — it's raw RPC + Helius `getTransactionsForAddress`. There's no precedent in the SDK to copy.
- These mints have **no collection/group key**, so `getAssetsByGroup` has nothing to query.

So a live DAS enumerator can't be written today without (a) a Helius/DAS RPC and (b) a grouping key. Shipping one blind would be unverifiable.

**What I did instead:** demoted `skills:index` from "registry of truth" to an explicit, rebuildable **CacheLayer** (`core/skillSource.ts` — exactly coding-info §⑤'s framing). The mint is now the source of truth: `supply` always hydrated from the mint, and `verifyTraits` re-reads category/hashtags from on-chain `TokenMetadata`. Search + reputation enumerate through a swappable `SkillSource`, so a DAS source drops in later with **zero caller changes**.

**The grouping key now exists for free:** after the protocol-minter change, every skill shares the **minter as mint authority**. A future `getAssetsByAuthority(minter)` enumerator is viable once a DAS RPC is wired (and, if DAS keys on update-authority, point that at the minter too via `createUpdateAuthorityInstruction`, which exists). **Decision needed:** which DAS provider / RPC, and confirm whether it keys on mint- or update-authority.

### 2. On-chain note gating — the Candy Guard / Path B question

Note/comment gates are currently **client-side only**. Root cause (verified in both the published SDK and the local clone):

- The contract gate resolves the holder's ATA with the **legacy** `TOKEN_PROGRAM_ID` in the seeds (`utils/ata.ts`). Skill mints are **Token-2022** → the holder's real ATA never matches → `resolveSignerAta` throws `missing signer_ata` on **every** write. A natively-gated notes table is therefore unusable; I left it open + client-guarded and documented why.

The proposed fix — **Metaplex Collection + Candy Guard**, leaning on the SDK's `GateType.Collection` — is a real candidate (the gate genuinely resolves a `metaqbxx` metadata PDA, so Metaplex collections were clearly contemplated). But three hard caveats before committing:

1. **MPL Core breaks the gate.** `GateType.Collection` reads a **legacy SPL ATA** (offset-0 mint) + an mpl-**token-metadata** PDA. Core assets have no ATA and aren't token-metadata → the gate can't see them. So "Candy Machine/Core + Core non-transfer plugin" is self-contradictory — the held asset **must be a classic Token-Metadata NFT**.
2. **Then soulbound is no longer free.** Token-Metadata has no `NonTransferable`; soulbound becomes pNFT + Token Auth Rules / freeze — heavier than today's one-line extension.
3. **The IQ Collection gate is unproven.** Zero tests exercise it; the code comment says *"for now the ATA mint is used"*; `signer_ata`/`metadata_account` are `optional` in the IDL. "Stronger than client-side" is a hypothesis until proven on devnet.

It also **reverses skill-nft-structure.md §5**, which explicitly rejected per-copy Metaplex mints ("500 copies = 500 print mints, heavy") in favor of Token-2022's free on-chain supply count. Candy = a new NFT per buy + popularity via DAS = the cost §5 chose to avoid.

**The decision reduces to one question: is on-chain enforced note/unlock gating a must-have now?**
- **No** → stay on Token-2022 (current state). Authority is already solved by the protocol minter; notes stay client-guarded, which is all the deployed contract can do for any non-legacy asset anyway.
- **Yes** → Path B is the only deployed-contract route, but commit knowing the rearchitecture: classic Token-Metadata (not Core), harder soulbound, per-skill Candy Machine, DAS for counting — and **finish + devnet-prove `GateType.Collection` first** (a throwaway spike against one real minted Token-Metadata NFT is the single fact that de-risks the whole call).

**Recommendation:** don't rearchitect the NFT engine on an unproven, self-described "for now" gate. Keep Token-2022; if on-chain note enforcement becomes a hard requirement, spike the Collection gate on devnet before adopting Path B.

zo-sol:
## 📋 T2 (on-chain skills) review — full picture + the umbrella-collection fix

Mapping out what T2 was meant to be → how this PR's modules wire together → how the NFT write/read flow works (and the missing read piece) → what to change (umbrella structure) → a simpler note row.

---

### 1. What T2 was (per the plans)

Per `roadmap-2tracks.md`, T2 = **the on-chain layer where a wallet (= an agent) publishes / searches / buys skills and leaves reputation notes.** The core design lives in `skill-nft-structure.md`, in one line:

> A skill = **one Token-2022 mint** (soulbound · `supply` = popularity · `uri` = code-in txid). **The collection IS the skill list** — do not build a separate registry table.

```mermaid
flowchart LR
    AG["wallet = agent"] --> PUB["publish<br/>code-in text + Token-2022 mint"]
    PUB --> SK["skill NFT<br/>(NonTransferable = soulbound)"]
    SK --> BUY["buy = mint 1<br/>supply++ = popularity"]
    SK --> SEARCH["search<br/>trait filter + sort by supply"]
    AG --> NOTE["notes<br/>(on-chain comments, holder-gated)"]
    SK --> REP["reputation<br/>= sum of supply over my skills"]
    style SK fill:#f0e6ff,stroke:#93c
```

Roadmap T2-1~T2-8 (core / publish / buy / search / notes / validation / workflow / mcp) are **all implemented in this PR** — the mapping is accurate. Nice work getting the whole track stood up.

---

### 2. How the modules wire together right now

Every arrow below is an actual import in the PR.

```mermaid
flowchart TB
    subgraph CORE["core/ — shared base, wraps the iqlabs SDK"]
        CHAIN["chain.ts<br/>codeIn · readCodeIn · writeRow · readRows<br/>ensureDbRoot · ensureTable · signerAddress"]
        SEED["seed.ts<br/>table hints + column lists"]
        TYPES["types.ts"]
        SRC["skillSource.ts<br/>listSkills — the 'skill list' entry point"]
    end

    subgraph NFT["nft/ — Token-2022, the only token module"]
        SKILL["skill.ts<br/>publishSkill · buySkill"]
        WF["workflow.ts<br/>publishWorkflow · unlockWorkflow"]
        T22["token2022.ts<br/>createSkillMint · mintSkillToken<br/>readSkillMintMetadata · getMintSupply"]
        MINT["minter.ts<br/>protocol minter, Path A authority<br/>env only, no SDK/core deps"]
    end

    subgraph VALD["nft/validation/ — wrapped chain, not compose"]
        VWF["WorkflowAdapter"] --> VON["OnchainAdapter = defaultValidator"]
        VON --> VST["StrictAdapter"] --> VCO["SkillsShCompatAdapter"]
        VPARSE["parse.ts, shared by all"]
    end

    subgraph NOTESM["notes/"]
        NOTES["notes.ts<br/>postNote · postAgentNote · readNotes"]
        BAL["balance.ts<br/>getBalance — spl-token only"]
    end

    SEARCH["search/search.ts"]
    REP["reputation/reputation.ts"]
    MCP["mcp/server.ts"]

    IDX[("skills:index table<br/>⚠️ de-facto registry")]

    %% publish paths
    SKILL -->|codeIn + writeRow| CHAIN
    SKILL -->|createSkillMint| T22
    SKILL -->|defaultValidator| VON
    SKILL -->|resolveMinter| MINT
    SKILL -. "writes 1 row" .-> IDX

    WF -->|codeIn + writeRow| CHAIN
    WF -->|createSkillMint| T22
    WF -->|defaultWorkflowValidator| VWF
    WF -->|"getBalance — prereq gate"| BAL
    WF -->|resolveMinter| MINT
    WF -. "writes 1 row" .-> IDX

    T22 -->|signerAddress| CHAIN
    T22 -->|"resolveMinter, tryMinterPubkey"| MINT
    VON --> VPARSE

    %% list source
    SRC -->|readRows| IDX
    SRC --> CHAIN

    %% consumers
    SEARCH -->|listSkills| SRC
    SEARCH -->|"getMintSupply, readSkillMintMetadata"| T22
    REP -->|listSkills| SRC
    REP -->|getMintSupply| T22
    REP -. "reads notes:skill for notesReceived" .-> CHAIN
    NOTES -->|"listSkills — agent-note gate"| SRC
    NOTES -->|"getBalance — holder gate"| BAL
    NOTES --> CHAIN
    MCP -->|searchSkills| SEARCH
    MCP -->|buySkill| SKILL

    style IDX fill:#fee,stroke:#c33,stroke-width:2px
    style SRC fill:#fff7e6,stroke:#ca0
    style MINT fill:#eef,stroke:#33c
    style VON fill:#e6ffe6,stroke:#3a3
```

**Reading the graph:**

- **Publish (`skill.ts` / `workflow.ts`)** is the orchestrator — it touches everything: ① `validation` gate → ② `chain.codeIn` to inscribe → ③ `token2022.createSkillMint` (which itself hands authority to `minter`) → ④ `chain.writeRow` into `skills:index`.
- **`minter.ts` is a leaf** — env-config only, no `core`/SDK imports. Both `token2022` (sets authority at publish) and `skill`/`workflow` (co-signs `mintTo` at buy/unlock) pull it.
- **Validation is a *wrapped chain*, not `compose()`** — `WorkflowAdapter` instantiates `OnchainAdapter`, which instantiates `StrictAdapter`, which instantiates `SkillsShCompatAdapter`; `parse.ts` is shared by all. `compose()` exists for *adding* the security-LLM adapter on top, but the default validators are this nested `new` chain.
- **`balance.ts` is the shared holder-gate primitive** — depends only on `@solana/spl-token` (no `core`). Both `notes.ts` (comment gate) and **`workflow.ts` (prerequisite gate)** reuse it — i.e. the nft module reaches *back into* notes/ for `getBalance`.
- **The `skills:index` table is the hub** — `skillSource.listSkills()` is the single read point, and `search` / `reputation` / `notes`(agent-note gate) all go through it. `reputation` additionally reads the `notes:skill:<id>` tables directly to count `notesReceived`. This is the registry the umbrella fix (§4) removes — and because every consumer sits behind `skillSource`, the swap stays contained to that one file.

---

### 3. How NFTs are written / read — and the missing read piece

**Write (publish):**

```mermaid
flowchart LR
    TXT["SKILL.md text"] --> CI["codeIn (SDK)<br/>on-chain inscription"]
    CI --> TXID["returns txid"]
    TXID --> MINT["createSkillMint<br/>TokenMetadata.uri = txid"]
    MINT --> M["skill mint (uri=txid, traits)"]
    style TXID fill:#cfc,stroke:#3a3
```

- SDK **`codeIn(...)`** sends the inscription tx and returns **the tx signature (txid)** (confirmed in `code_in.js`: it ends with `return signature`).
- That **txid is stored in the NFT `uri`** instead of an http url (the IQ6900 pattern). ✅ as intended.

**Read (the intended two steps):**

```mermaid
flowchart LR
    MADDR["skill mint address"] --> RM["readSkillMintMetadata<br/>(getTokenMetadata)"]
    RM --> URI["recover uri = txid"]
    URI --> RC["readCodeIn(txid)<br/>(SDK)"]
    RC --> BODY["SKILL.md body text"]
    style RC fill:#fee,stroke:#c33
    style BODY fill:#fee,stroke:#c33
```

- Step 1 (mint → uri): `readSkillMintMetadata` exists ✅
- Step 2 (txid → text): the `readCodeIn(txid)` wrapper exists, and its signature matches the SDK's actual `{metadata, data}` ✅
- **The round-trip key (txid) lines up correctly** — what `codeIn` returns is exactly what `readCodeIn` consumes.

**🧩 Missing puzzle piece (read side is incomplete):**

> There's **no `getSkillText(mint)`-style function joining the two steps, and `readCodeIn` is never called anywhere in the PR.** So the path to actually read a published skill's body back is empty. `search.ts` reads name/description from the index table, never the body.
>
> Analogy: the letter is dropped in the mailbox (`codeIn`), the postcode is written on the NFT (`uri = txid`), and the machine that fetches a letter by postcode exists (`readCodeIn`) — but there's no "show me this NFT's letter" button.
>
> Also: chunked (>700B) inscription round-trips are never verified against a real chain (tests are all mocked). Only inline (≤700B) is effectively exercised.

→ **Suggestion:** add one `readSkillText(mint): Promise<string>` that composes `readSkillMintMetadata` + `readCodeIn`, and export it from the barrel so the skill-detail / search views can call it.

---

### 4. 🔴 To fix: there's no umbrella (collection) — use a collection, not a table

Right now the skill mints are **strangers with no link to each other**, so the chain can't be asked "give me all our skills." That's why publish writes a row into `skills:index` and we use that as the list — which is exactly the registry table `skill-nft-structure.md §2` ⚠️ says **not** to build, and it has a real weakness: **a mint whose row didn't get written is invisible to search.**

**Fix = two umbrellas (a skill collection + a workflow collection) as TokenGroups, and enroll each mint into the right one at publish.** (An "umbrella" is **not one big NFT** — it's a **nameplate that groups items**; you don't buy *into* it, you use it to **enumerate** "everything under this nameplate.")

```mermaid
flowchart TB
    subgraph SKU["☂️ Skill collection (TokenGroup)"]
        SA["skill A mint<br/>supply 50"]
        SB["skill B mint<br/>supply 9"]
    end
    subgraph WFU["☂️ Workflow collection (TokenGroup)"]
        WA["workflow X mint<br/>requiredSkills=[A,B]"]
    end
    DAS["DAS getAssetsByGroup<br/>= enumerate everything in a collection"]
    SKU --> DAS
    WFU --> DAS
    DAS --> SRC["skillSource.listSkills<br/>from here, not the table"]
    style SKU fill:#f0e6ff,stroke:#93c
    style WFU fill:#e6f0ff,stroke:#36c
    style DAS fill:#cfc,stroke:#3a3
```

**Why two umbrellas / what to change:**

| What | Now | Change to |
|---|---|---|
| Collection nameplate | none | create **two TokenGroup mints** once at bootstrap (`SKILLS_COLLECTION`, `WORKFLOWS_COLLECTION`) |
| Minting (`createSkillMint`) | NonTransferable + MetadataPointer + TokenMetadata | **add `GroupMemberPointer` + TokenGroupMember** → enroll into a collection. publishSkill → skill umbrella / publishWorkflow → workflow umbrella |
| List source (`skillSource.ts`) | `readRows(skills:index)` | swap to **DAS `getAssetsByGroup(umbrella)`** (interface stays the same — just `indexTableSource` → `dasSource`) |
| skill vs workflow split | table `type` column | **by which umbrella** (search's `type` filter becomes "which collection to enumerate") |
| `skills:index` table | 5 modules depend on it | **droppable** once DAS enumeration lands (keep as a rebuildable cache until then) |

> ⚠️ Token-2022 caveat (`skill-nft-structure.md §1`): a mint being *both* group and member is undocumented — don't rely on it. So **the umbrella mint (group) and the item mints (members) must be separate.** Per-skill counts still come from each mint's **`supply`**, not the umbrella's `size` (keeps the current supply model unchanged).

**Files to touch:** `token2022.ts` (add the extensions) · `seed.ts` (two collection constants) · the bootstrap script (create the umbrellas once) · `skillSource.ts` (DAS source). **Consumers (search / reputation / notes / mcp) don't need to change** — they sit behind `skillSource`.

---

### 5. 📝 Note row — let's make it simpler

`notes.md` specs the row as `{author, body, attach:{kind,url}, ts}`; the code uses `{id, author, subject, text, gitLink, isSelfNote, timestamp}` — **both carry dead weight.** Trimming on the "keep it simple" principle:

- Drop `subject` → **the table key already is the subject** (`notes:skill:<mint>` / `notes:agent:<wallet>`).
- Drop `isSelfNote` → **derive from `author == subject`** (notes.md §3 also says "no flag, derive from author").
- Flatten `attach:{kind,url}` → **a single URL string** (on-chain vs off-chain git is obvious from the URL).
- One free-form `meta?` for future/experimental fields (a waiting room before a field earns promotion).

**Proposed shared row (same shape for skill notes and agent notes):**

```jsonc
{
  "id":      "<author>:<ts>",   // table PK
  "author":  "<base58>",
  "text":    "body",
  "gitLink": "https://...",      // optional attachment
  "ts":      1700000000,
  "meta":    { }                 // optional, any extra data
}
```

7 fields → 4 core + 2 optional. Reused as-is for skill / agent / (future) audit comments. (For `id`, mix in a counter/nonce to avoid same-ms collisions.)

zo-sol:
Pushed a small rename (`4ad627e`) — no logic changes, just naming:

- `src/mcp/` → `src/skill-market/`
- `skill-market/server.ts` → `skill-market/index.ts`

**Why:** the `mcp/` name describes the *transport* (Model Context Protocol), not what the module *is* — the agent-facing entry into the on-chain skill marketplace. A couple of us read it as "claude/codex MCP config" at first glance rather than "the skill market's gateway", so the name was hiding the module's purpose. `index.ts` because the file is the folder's entry point (a tool index + dispatcher over `searchSkills`/`buySkill`), not a generic "server".

Tool/function names (`search_skills`, `buy_skill`, `createAgentMcpServer`) are left **unchanged** — those are the MCP-protocol names the engines actually call, and they're already clear. 118 tests pass, build clean.

Still studying the rest of this PR (the nft / search / reputation / validation layers) before proposing how it wires into the runtime + surfaces — will follow up with the integration plan separately. Nothing else here is touched for now.

zo-sol:
## Workflow gate contract — `agent-workflow-nft` (built + deployed to devnet)

Standard Token-2022 can't express a **conditional** mint on a mint, so workflows (and **only** workflows — skills are still bought freely) now go through a small Anchor program. The workflow mint's authority is a program PDA, so the *only* path to a workflow token is `buy_workflow`, which checks the prerequisites on-chain first — a raw `mintTo` can't be signed by anyone.

**Repo:** https://github.com/IQCoreTeam/agent-workflow-nft (Anchor 0.32.1)
**Devnet program:** `3ptXj4yuaQG51WTA3SZZ37jGvYFgMhgXnSKWJLASJNkt`
**Skills collection it gates against:** `4exdqNEcXixiMzenEBts2cE7qLmMvcVtHCjsZUGBm4Gt` (`constants.rs`, swap before mainnet)

```mermaid
flowchart TB
    subgraph Publish["publish_workflow (register, once)"]
        P1["store required_skills in a config PDA"]
        P2["verify each skill mint is in the official collection<br/>(read its TokenGroupMember.group — O(1), no scan)"]
        P3["reject duplicate skills"]
        P1 --- P2 --- P3
    end
    subgraph Buy["buy_workflow (gate + mint)"]
        B1["read required_skills from the config PDA"]
        B2["for each skill token acct: Token-2022-owned,<br/>right mint, owned by buyer, amount >= 1"]
        B3{"holds all?"}
        B4["pay creator (if priced)"]
        B5["mint 1 workflow token via mint-authority PDA"]
        B1 --> B2 --> B3
        B3 -->|no| BX["reject (revert)"]
        B3 -->|yes| B4 --> B5
    end
    Publish -.->|config PDA| Buy
    style Buy fill:#efe,stroke:#3a3
    style BX fill:#fee,stroke:#c33
```

**Design notes**
- **Membership is O(1) per skill** — each skill mint carries its own `TokenGroupMember.group`; we read that one field, never enumerate the collection. Cost doesn't grow with the catalog.
- **Collection check happens at publish (once)** since mints are immutable — keeps `buy_workflow` cheap (no per-purchase re-check).
- **Anti-spoof:** each skill token account must be owned by the Token-2022 program (a forged look-alike can't pass), and `required_skills` lives in the config PDA (can't be forged by the client).
- `unlock` → **`buy_workflow`** (publish + buy instructions).

**Verified end-to-end on devnet** (`tests/workflow-gate.ts`, 4 passing): a holder of all required skills buys successfully; a wallet missing one is rejected on-chain.

**Next:** wire the SDK's `unlockWorkflow` to call `buy_workflow`, and move the workflow mint authority to the program PDA at publish (so the client-side balance pre-check can be dropped — the chain enforces it now). Plan doc updated: `plans/workflow-nft.md` §2.

zo-sol:
## Update — the gate now covers skills too (unified into one `item` model)

While wiring the SDK, it turned out **skills and workflows are the same shape**, so the gate program was generalized: `publish_workflow`/`buy_workflow` → **`publish_item`/`buy_item`**. An item's `required_skills` is:
- **empty → a skill** (anyone buys; the gate loop runs zero times)
- **filled → a workflow** (buyer must hold every listed skill)

### Why skills go through it too: no more minter keypair
Token-2022 `mintTo` requires the **mint authority to sign** (confirmed against the docs + our own `resolveMinter` which throws without a secret). With the old design the protocol-minter **keypair** had to co-sign every skill buy — which means either the secret sits in the browser (anyone could mint infinitely) or every buy routes through our server (not trustless).

Now every item mint's authority is the program's **mint-auth PDA**, so:
- **no minter secret anywhere** — the program signs the mint under its PDA
- price + creator are read from the item's **config PDA** (set at publish), so the client can't forge them. `BuySkillInput` dropped `price` / `iqFeePercent` / `minter`.

### Status
- Contract (separate repo, redeployed to devnet, same program id `3ptXj4...`): `WorkflowConfig`→`ItemConfig`, seed `"workflow"`→`"item"`, empty `required_skills` allowed. **4 integration tests pass** (skill-holder buys; missing-skill rejected on-chain).
- SDK: `buySkill`/`publishSkill`/`publishWorkflow`/`unlockWorkflow` all call `buy_item`/`publish_item` via raw web3.js (`packages/core/src/nft/workflowGate.ts`) — no anchor dep. **82 tests pass, tsc clean, build green.**

Plan doc `plans/workflow-nft.md` §2 updated with the unified flow + mermaid.

zo-sol:
Superseded by #11 — the Step-0 core layer was relocated into packages/core/ and merged to main (commit b159112, with mega123-art credited as co-author). Closing this branch's PR; the original commits remain on feat/step0-core-layer.

PR #56, opened 2026-06-19
Mobile Skill Market Parity, Workflow Selling, and Plugin NFT Plumbing

mega123-art:
# PR Description: Mobile Skill Market Parity, Workflow Selling, and Plugin NFT Plumbing

This Pull Request brings full mobile webview parity for skill management (removing/re-equipping), enables workflow selling from the marketplace screen, improves agent passive skill awareness, and merges the on-chain plugin installation plumbing.

---

### 1. Mobile Skill Market Parity (Equip / Dispose)
- **State Integration**: Wired up the `marketDisposed` state record in the webview store to sync with SSE server payloads.
- **UI Parity**: 
  - Faded cards representing un-equipped (disposed) skills with a dashed border and an "un-equipped" badge in the catalog grid.
  - Added "Remove Skill" and "Re-equip Skill" buttons to the `SkillDetailView` footer with direct RPC message triggers (`disposeSkill` / `reEquipSkill`).

### 2. Workflow Selling
- **Publish Parsing**: Added dynamic frontmatter parsing to the `publishSkill` RPC callback. If a publish input contains `type: workflow`, it automatically parses prerequisites (`requiredSkills`) and delegates the minting to `publishWorkflow` instead of `publishSkill`.

### 3. Agent Skill Awareness
- **Instruction Mapping**: Updated `updateSkillsSection` to pass the correct active runtime engine (Claude vs Codex).
- **Paths & Prompts**: Guide agents with precise path mappings (`~/.claude/skills/` or `~/.codex/skills/`) and instructions on using standard file tools (like `view_file` on `SKILL.md`) to read skill bodies.

### 4. Plugin NFT Collection & Installation Support (Merged)
- **NFT Collection Plumbing**: Integrated collection schemas and DAS/Indexer mapping logic for plugin NFTs.
- **Engine-specific Installation**: Resolves plugin manifests and implements automated installation for the target engine (Claude/Codex).
- **Detail View & Installs**: Exposes plugin installation action triggers in `SkillDetailView`.

---

## Merge Resolution & Integration

During the merge of `plugin-nft-collection-plumbing` into `mobile-skill-market-parity`, we resolved conflicts in:
- `protocol.ts`: Combined schemas for client-side messages (`disposeSkill`, `reEquipSkill`, `installPlugin`) and server responses (`disposeResult`, `reEquipResult`, `pluginInstallResult`).
- `store.tsx`: Merged reducer handlers to support both skill status changes and plugin installation alerts.
[...]

PR #62, opened 2026-06-20
Point passive skill awareness at active runtime

mega123-art:
## Summary
- read installed skills from the active runtime directory instead of always Claude's dir
- include the concrete Claude/Codex SKILL.md path in the managed awareness block
- update the unit test to verify Codex reads ~/.codex/skills separately

## Verification
- pnpm --filter @iqlabs-official/agent-sdk exec tsc --noEmit
- pnpm --filter @iqlabs-official/agent-sdk exec vitest run src/memory/skillsSection.spec.ts

Split out of #56 per review feedback.

### DIFF
--- a/packages/core/src/memory/skillsSection.spec.ts
+++ b/packages/core/src/memory/skillsSection.spec.ts
@@ -5,9 +5,10 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
 // replaced in place, and never disturbs content outside the markers.
 const files = new Map<string, string>();
 let skillDirs: string[] = [];
+let codexSkillDirs: string[] = [];
 
 vi.mock("node:fs/promises", () => ({
-  readdir: vi.fn(async () => skillDirs.map((name) => ({ name, isDirectory: () => true }))),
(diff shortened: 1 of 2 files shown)


zo-sol:
Thanks @mega123-art — great change. Making the skills block read the *active* runtime's dir (`~/.codex/skills` vs `~/.claude/skills`) and pointing the memory line at the matching `SKILL.md` path is exactly right, so the agent only reaches for skills the running engine can load.

I cherry-picked the core of this onto `main` in c1a9f85 (with you as co-author), keeping your `skillsSection.ts` + spec as-is (specs pass). I took only the core file rather than merging the branch because this PR predates the market RPC refactor on `main` (`3b6b601`) and the dispose/re-equip work (#60) — a full merge would have silently reverted the `attachMarketHandlers` extraction (verified via a trial merge: `attachMarketHandlers` count → 0). The webview-side bits here are superseded by what's already on `main`.

Closing as **cherry-picked into main**. Really appreciate it — credit's on the commit. 🙏

PR #63, opened 2026-06-20
Fix mobile market owned skills payload crash

mega123-art:
## Summary
- send the localhost/mobile ownedSkills event with the expected names array instead of spreading an array into an object
- include owned and disposed mint maps so the payload matches the shared market contract
- guard the React reducer against malformed ownedSkills events so marketOwned never becomes undefined

## Verification
- pnpm --filter agentnet-localhost build
- pnpm --filter agentnet-webview build

This targets the mobile market black-screen crash path when entering the market.

### DIFF
--- a/surfaces/localhost/src/index.ts
+++ b/surfaces/localhost/src/index.ts
@@ -467,8 +467,12 @@ function attachChat(id: string, c: Client, rt: AgentRuntime) {
       }
       case "ownedSkills": {
         try {
-          const r = await mkt.ownedSkills();
-          c.send({ type: "ownedSkills", ...r });
+          const names = mkt.ownedNftSkills ? await mkt.ownedNftSkills() : await mkt.ownedSkills();
+          const [mints, disposedMints] = await Promise.all([
+            mkt.ownedSkillMints ? mkt.ownedSkillMints().catch(() => ({})) : Promise.resolve({}),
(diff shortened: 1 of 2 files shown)


zo-sol:
Thanks a lot for this, @mega123-art — you diagnosed a real crash precisely. `ownedSkills()` returns a `string[]`, so `c.send({ type: "ownedSkills", ...r })` spread the array into `{0:..,1:..}`, dropping `names` and leaving `marketOwned` undefined — which blew up the reducer (the mobile market black screen). Spot on. 🙏

The reason I'm closing rather than merging: the exact same fix already landed on `main` in commit `3b6b601` ("Fix mobile market RPC setup"), on both sides:
- server: `ownedSkills` now sends `{ names: await mkt.ownedNftSkills(), mints, disposedMints }` instead of spreading
- reducer: `marketOwned: Array.isArray(ev.names) ? ev.names : []`

We verified it on a Samsung device today — the market now loads agents + skills with no crash.

Because this PR branch predates `3b6b601`, merging it now would actually regress `main`:
- it would drop the dispose/re-equip feature merged in #60 (`disposeResult` / `reEquipResult` cases and the `marketDisposed` state)
- it would revert the `attachMarketHandlers` extraction (which also makes the market reachable from the onboarding/storage client)

So your fix is effectively already in — closing as **already incorporated**, with credit to you for catching it. Really appreciate the sharp eye, and please keep them coming! 🚀

Issue #115, opened 2026-07-12
Run the engine natively (no proot in the hot path): structural fix for the untrusted_app transport corruption


zo-sol:
> **STATUS UPDATE (2026-07-12):** the T0 probe (comment below) killed this issue's core premise — the Ubuntu **glibc rootfs cannot be launched through the bionic linker** (non-PIE rejected; glibc rtld aborts on PHDR under bionic execve; even PIE glibc binaries segfault). The linker-chain modules below (T1/T2 as written) are dead on arrival. The revised plan of record is in [this comment](https://github.com/IQCoreTeam/AgentNet/issues/115#issuecomment-4951076737): **Phase 1 — instrument and patch proot's transport under `untrusted_app` (timeboxed); Phase 2 — if unfixable, adopt a bionic (termux-ecosystem) userland for the engine.** The in-app probe (`NativeProbeActivity`) stays as the standing on-device gate. Original text kept below for the record.

## Goal

Eliminate the #112 bug class at the architecture level: run the four engine binaries (node, claude, codex, git) as native Android processes, with no ptrace layer between them and the kernel. The dulwich clone shim from #114 is an interim patch for one subcommand of one tool; this issue is the structural fix, and it must hold on Snapdragon and MediaTek devices and in the Play Store release build (targetSdk 35).

Context: #112 (root cause + verification threads), #114 (interim shim, merged as `509dee6`), fix ladder: https://github.com/IQCoreTeam/AgentNet/issues/112#issuecomment-4950397541

## Why the current architecture breaks

Every engine syscall crosses proot's ptrace translation. Under the `untrusted_app` SELinux domain (targetSdk 35, required for Play), that translation corrupts multi-process piped/socket data — measured as truncated pack streams and TLS decode failures. Anything pipe-heavy (git today, npm/cargo tomorrow) is exposed.

```mermaid
flowchart TD
    WV["WebView UI"] --> SM["ServerManager"]
    SM --> GE{{"GuestExec seam"}}
    GE --> DPE["DirectProotExec"]
    DPE --> PROOT["proot (ptrace tracer)"]
    PROOT -->|"every syscall translated"| NODE["node server"]
    NODE --> CLAUDE["claude CLI"]
    NODE --> CODEX["codex CLI"]
    NODE --> GIT["git + git-remote-https + index-pack"]
    PROOT -. "CORRUPTS multi-process<br/>pipe/socket data under<br/>untrusted_app (#112)" .-> GIT

[...]

PR #116, opened 2026-07-13
fix(android): #115 root cause — drop proot --link2symlink (fixes git, pnpm, all link() tools)

mega123-art:
## What

Fix the #112/#115 git object-loss bug at its root: write a system `/etc/gitconfig` into the guest with `core.createObject = rename`, so git stops using the `link()` syscall that proot silently breaks.

**One line of effect** ([`Installer.kt`](../blob/fix/115-git-object-rename/surfaces/android/app/src/main/java/com/iqlabs/agentnet/Installer.kt)):

```kotlin
writeFresh(File(p.rootfs, "etc/gitconfig"), "[core]\n\tcreateObject = rename\n")
```

## Why

Under the `untrusted_app` SELinux domain (targetSdk 35, required for Play), **proot's `link()` is a false success** — it returns `0` but the target file never appears. git's default loose-object finalize path (git 2.43 `object-file.c`) is `mkstemp → fsync → link(tmp,final) → unlink(tmp)`; git trusts the `0` from `link`, reports the object written, and **the object is silently lost**. This surfaces as `fatal: remote did not send all necessary objects` / `bad object` on `git clone`, and would hit `commit`/`fetch`/`pull` the same way.

`core.createObject=rename` sets `OBJECT_CREATION_USES_RENAMES`, which makes git **skip `link()` entirely** and use `rename()` — which proot handles correctly.

## Proof (on-device, headless, `untrusted_app` — SM-A356E / Android 16)

**Root cause caught at the syscall (probe9):**
```
link ret=0 errno=0 target_exists_after=0   tmp_obj_mvrgkg -> objects/f0/6b470c...
hash-object exit=0        # git thinks it worked
count=0                   # nothing on disk
```

**Fix proven (probe10):**
```
link mode  (default):            50 written, survivors_on_disk=0    ← bug reproduced
rename mode (core.createObject):  50 written, survivors_on_disk=50   ← fixed
```

**Verified through the shipping path** (system `/etc/gitconfig`, plain `git clone`, no per-command flag):
```
git config --show-origin core.createObject  ->  file:/etc/gitconfig  rename
git clone https://github.com/expressjs/express.git  ->  exit 0, done.
HEAD=ae6dd376   in-pack: 53322   packs: 1   checkout: OK   worktree files: 213   fsck errors: 0
```

## Relation to other work

- **Supersedes #114** (the dulwich clone shim, merged `509dee6`): that fixed `clone` only, by swapping real git for pure-Python dulwich (which writes objects with create+rename, dodging `link`). This PR fixes **all** git commands with **real git** and one config line. Once merged, #115 T3 (remove the dulwich shim) can proceed.
- **Root-cause hunt:** builds directly on @zo-sol's handoff in #115 (proven eliminations: network, socket recv, rseq, l2s, rename; narrowed to git's loose-object write). Full evidence: [#115 comment](https://github.com/IQCoreTeam/AgentNet/issues/115#issuecomment-4951631664) and the confirmation [#115 comment](https://github.com/IQCoreTeam/AgentNet/issues/115#issuecomment-4955044987).
- **Original context:** #112 (root cause + verification threads), #115 (structural-fix issue).
- **Not this PR (still open):** the *basement* fix — patching proot's `linkat` translation so `link()` stops lying for **every** tool (npm/cargo/…), in our own proot build (we own it from the Seeker process_vm fix; GPLv2 → publish the patch). `core.createObject=rename` fixes git, the only proven-affected tool today; the proot patch stays tracked in #115.

## Also in this branch (debug kit, not shipped code)

- `debug/issue-115/probes/probe10.sh` — the fix test (reproduce in link mode → prove in rename mode → real clone).
- `debug/issue-115/shims/liblinkfix.c` — a broader `LD_PRELOAD` option (returns `EXDEV` on object-path `link`/`linkat`, forcing the caller's own rename fallback) for any **non-git** tool that hits the same false-success and has no config knob.
- `debug/issue-115/README.md` — the fix ladder (config → shim → proot patch).

> Headless reproduction used for all the evidence above lives on `feat/115-native-exec-probe` (`NativeProbeActivity` guest-script mode: `am start … -e guestScript /root/probeN.sh` runs a probe through proot **from the app process**, so git inherits `untrusted_app` — same domain as the in-app chat, drivable over pure adb, no agent auth). Kept separate from this shippable fix.

## Test plan

1. Fresh install (or clear app data so the installer re-runs `configureGuest`).
2. In-app: `git clone https://github.com/expressjs/express.git` → completes, `git fsck` clean.
3. Confirm the config is live: guest `git config --show-origin core.createObject` → `file:/etc/gitconfig  rename`.

### DIFF
--- a/debug/issue-115/README.md
+++ b/debug/issue-115/README.md
@@ -38,6 +38,30 @@ Decision table is in the handoff comment (link at top). Short version: if the lo
 `link ret=0 ... target_exists_after=0`, proot's `link` is a false success — that is the
 root cause.
 
+## Candidate fix — run probe10 right after probe9 confirms link
+
+git 2.43 `finalize_object_file()` (object-file.c, verified against v2.43.0): with
+`core.createObject=rename` (`OBJECT_CREATION_USES_RENAMES`) it **skips `link()` entirely** and
+uses `rename()`, which works under proot (probe6). So the interim fix is **one line of git
+config — no shim, no proot rebuild**:
+
+```
+run sh /root/probe10.sh and save the full output to /root/probe10_result.txt
+```
+probe10 reproduces the loss in link mode, shows it gone in rename mode, then does a real
+network clone with the fix. Verdict: fixed iff rename-mode survivors=50 AND clone exit=0.
+
+**Fix ladder (apply in order once probe9 names `link`):**
+1. **Config, git-only, zero code** — `core.createObject=rename` in the guest (system gitconfig
+   or `ServerManager.buildGuestEnv` via `GIT_CONFIG_*`). Fixes the shipping bug (git is the only
+   proven-affected tool). Ship this.
+2. **`shims/liblinkfix.c`, all tools** — LD_PRELOAD that returns EXDEV on object-path `link`/
+   `linkat`, forcing the caller's own rename fallback. Use only if a non-git tool (npm/cargo)
+   hits the same link false-success and has no config knob.
+3. **Basement, proot `linkat` patch** — fix the false-success in our own proot build (we own it
+   from the Seeker process_vm fix; GPLv2 → publish the patch). The true fix: covers every tool,
+   no per-tool workaround. This is the long-term "fix the basement" answer.
+
 ## How to stage these on a device
 
 ```bash
--- a/debug/issue-115/probes/probe10.sh
+++ b/debug/issue-115/probes/probe10.sh
@@ -1,0 +1,41 @@
+# issue #115 — CANDIDATE FIX test. Gated on probe9 first confirming `link ret=0
+# target_exists_after=0` for .git/objects (proot's link() is a false success).
+#
+# git 2.43 finalize_object_file() (object-file.c): if core.createObject=rename
+# (OBJECT_CREATION_USES_RENAMES) it `goto try_rename` and SKIPS link() entirely, using
+# rename() — which works under proot (probe6: 50 renames survive, 50 link-writes vanish).
+# So the interim fix is one line of git config: no LD_PRELOAD shim, no proot rebuild.
+# probe10 proves it end to end: reproduce the loss in link mode, show it gone in rename mode,
+# then a real network clone with the fix.
+#
+# Run as a TRUE app child (paste into the in-app agent chat), never run-as:
+#   run sh /root/probe10.sh and save the full output to /root/probe10_result.txt
+# then: adb shell "run-as com.iqlabs.agentnet cat files/rootfs/root/probe10_result.txt"
+set -u
+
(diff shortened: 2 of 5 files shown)


mega123-art:
## Update: found the real root cause — it's `--link2symlink`, and the fix is to remove it

The `core.createObject=rename` gitconfig (first commits) fixes git, but git isn't the only victim. Dug to the actual root cause; it's one flag, and removing it fixes **every** tool at once.

### Root cause (proven on-device, SM-A356E, untrusted_app)

In the `untrusted_app` domain the kernel **denies hardlinks**: `ln a b` → `Permission denied`, `links=1`. We launch proot with `--link2symlink`, whose extension catches that denial and **fakes success** — it pokes the syscall result to `0` and swaps in a symlink-refcount scheme (moving the file into `PROOT_L2S_DIR`, replacing the target with a symlink). That fake defeats callers' own error handling:

- git `finalize_object_file` falls back to `rename` **only when `link` fails** — never when it returns a lying `0`.
- pnpm's store-linker falls back to `copy` **only when the hardlink fails** — never when it lies.

So both get a `0`, skip their fallback, and the symlink dangles → silent git object loss and broken pnpm installs. This is exactly probe9's `link ret=0 target_exists_after=0`.

### The fix: drop `--link2symlink`

Then `linkat` fails **honestly** with EACCES and every tool with a fallback recovers itself. Verified on-device with l2s off:

```
ln a b:                     Permission denied  (honest — no more fake success)
git (forced link mode):     50/50 loose objects survive   (rename fallback)
pnpm (default hardlink):    589/589 files, require OK      (copy fallback)
node server boot:           server ready (HTTP 200)        (no regression)
```

One flag removed fixes git + pnpm + any link()-using tool — **no proot rebuild, no per-tool config**. This is the "fix the basement" answer: the defect was proot faking a syscall, and we stop it faking.

### What this PR now contains

1. **`DirectProotExec`: remove `--link2symlink`** — the root-cause fix, all tools.
2. **`Installer`: `core.createObject=rename` system gitconfig** — kept as belt-and-suspenders for git (skips the doomed link attempt entirely), and written on every launch so existing installs get it without a rootfs re-extract.

Supersedes the clone-only dulwich shim (#114) — real git, all commands, plus pnpm and the rest. #115 T3 (remove the dulwich shim) can proceed.

### Note on the earlier "l2s isn't it" elimination

The handoff eliminated `--link2symlink` because `ln` showed `links=2` (a real hardlink). That test was in `runas_app`, where hardlinks are **allowed** — so l2s passed the real link through untouched and looked innocent. The bug only appears in `untrusted_app`, where the real link is **denied** and l2s's fake-success path activates. Same reason the whole bug is domain-specific.

PR #143, opened 2026-08-12
Add an in-app PREVIEW tab: see and test what you build, inside the app

RemilioNubilio:
# Add an in-app PREVIEW tab: see and test what you build, inside the app

A fifth NAV_DOCK tab that frames a local dev server (`127.0.0.1:PORT`) in an iframe — the way VSCode's Simple Browser works. When the agent builds a web app, you see it live without leaving the app.

- **PREVIEW tab** (`PreviewScreen.tsx`) — a URL bar (port normalization, reload, open-external, safe-area top inset) over an iframe. Direct cross-origin iframe, **no proxy**: it's all one device's loopback, so the frame reaches the dev server directly and its own HMR reloads it for free (a path-prefix proxy would break absolute asset paths and HMR websockets). The tab stays mounted so a running dev server keeps rendering as you page away, and it sits outside the unlock gate.
- **Announce channel** (`preview.ts` + route in `index.ts`) — loopback-only `POST /preview/announce` (`{"port":N}` to set, `{"port":null}` to clear) validates the port and broadcasts `previewStatus` to every connected client; a `getPreviewStatus` RPC covers an announce that lands before the tab first mounts.
- **Tap-to-preview** (`Markdown.tsx` + `App.tsx`) — a loopback link the agent prints in chat becomes a one-tap "open preview": a `CustomEvent` bridge switches the shell to the tab and the panel frames the URL, instead of navigating the whole webview away.
- **Teach-the-agent** (`previewSection.ts`) — a managed memory block (mirrors the skills block) tells the agent to serve web apps in the background and hand you the `http://localhost:PORT` link. Gated on `AGENTNET_PREVIEW_PORT`, which only the surface that actually has the tab sets — CLI/VSCode agents are never told about an endpoint that doesn't exist.
- **Review fixes (this pass)** —
  - *Self-framing guard*: chat intercepts only absolute `http(s)` loopback links and skips same-host ones, and `normalizePreviewUrl` / `validatePreviewPort` both reject the app's own port — the app can never frame itself.
  - *CSRF guard on `/preview/announce`*: `isLoopback` can't stop a malicious page in an on-device browser (its fetch also arrives from loopback), but a browser always sends `Origin` — any Origin that isn't our own is rejected; agents (curl / node fetch) send none and pass untouched.
  - *Design tokens*: preview CSS uses the app's `--an-*` tokens; the one raw `#fff` (the frame backdrop dev pages assume) is deliberate and commented.
  - *De-dup*: `spliceIntoFile` / `memoryFileFor` hoisted out of `skillsSection.ts` and shared by both section updaters, with a write-guard so an empty block never creates or dirties a memory file that doesn't contain our markers.

Verified end-to-end on device: "build me a coffee-shop landing page" → the agent built it, served it in the background, and announced it — with no instruction to serve or preview. `previewStatus` broadcast, and the framed URL returned the rendered page (HTTP 200).

Scope note: this adds a whole surface, and that's product direction — yours to own. Treat it as a proposal: the tab, tap-to-preview, and the agent-teaching block are separable commits if you'd rather take part of it, and if an in-app preview isn't where you want the app to go, closing it is completely fine.

## Relates to

No linked issue — standalone proposal. Area: `surfaces/localhost` (mobile app shell) — NAV_DOCK / webview UI, chat-server routes, managed memory sections.

## Risks

**Medium.** Adds a new UI surface and a new loopback HTTP endpoint. Mitigations are in the diff: the endpoint is loopback-only with an Origin-based CSRF guard, self-framing is rejected at both validation points, and the agent-facing memory block is gated on `AGENTNET_PREVIEW_PORT` so CLI/VSCode surfaces are untouched. Purely additive — no existing tab or route changes behavior.

## Background — what & why

The agent can already build and serve web apps on-device, but the only way to see the result was to leave the app for an external browser — losing chat context and the tap-to-open flow. This adds a Preview tab (VSCode Simple Browser pattern) plus an announce channel so the agent's served app appears in-app, one tap from the chat message that produced it.

Start in `surfaces/localhost` — `preview.ts` (validation + payload, unit-tested standalone) and `PreviewScreen.tsx` (UI).

1. Run the localhost surface on device; note the 5th (PREVIEW) tab in the dock.
2. Manual path: `python3 -m http.server 8090` then `curl -X POST http://127.0.0.1:<app-port>/preview/announce -d '{"port":8090}'` → Preview tab frames `http://127.0.0.1:8090`.
3. Agent path: ask "build me a landing page" — the agent serves it in the background and announces it unprompted (the memory block at work).
4. Tap the loopback link the agent prints in chat → shell switches to the Preview tab framing that URL.
5. `npm test` in `surfaces/localhost` → 5/5 `preview` unit tests; `tsc --noEmit` clean on core.

---

### Code quality
Held to a strict bar — verified, not asserted:
1. **No meaningless wrappers with identical input/output** — no pass-throughs: `updatePreviewSection` adds the env gate + never-throw, `previewStatusMsg` derives the payload (`url` from `port`), and the review pass hoisted the existing splicer rather than wrapping it. ✅
2. **Scan existing functions first and reuse; never two functions with the same purpose** — one marker-block splicer in the codebase: `spliceIntoFile` / `memoryFileFor` are exported from `skillsSection.ts` and both the skills and preview updaters call them; the UI reuses `openExternalUrl`, the store, and `.an-page`, and the announce route reuses the existing per-client send loop. ✅
3. **No type variables that won't be reused; inline them** — `previewStatus` / `getPreviewStatus` are inline members of the existing protocol unions, `preview: { port, url }` is inline in `State`, and `validatePreviewPort`'s `number | null | false` lives at the signature; the only named type touched is the existing `TabKey` union, extended. ✅
4. **No non-essential parameters** — `validatePreviewPort(raw, serverPort)`: both load-bearing (the server's own port can't be a target). `spliceIntoFile` gained `start`/`end` only at the moment a second caller with different markers existed. ✅
5. **Keep responsibilities separated; if the design feels wrong, say so** — pure validation/payload logic in `preview.ts` (unit-tested without booting the chat server), the route + broadcast in `index.ts`, UI in `PreviewScreen.tsx`, the memory hint in `previewSection.ts`; the two deliberate oddities (no proxy; one non-token `#fff`) are called out in comments at the seam. ✅

`tsc --noEmit` clean on core; 5/5 new `preview` unit tests pass (`npm test` in `surfaces/localhost`). No regressions — the 8 failing vitest specs (rpc / seed / dasSource / skillSource / session) are pre-existing on main and env-dependent. Merges cleanly into main with zero conflicts.

## Evidence

Everything below is from a real on-device run of this branch; a reviewer can confirm the feature works from these artifacts without reading code.

**Before/After screenshots**

Before: the stock 4-tab dock on main (no capture — this PR adds a new fifth surface; "before" is simply its absence).

After:

5-tab dock after restart
The dock now shows five tabs — PREVIEW added — on a fresh app start.

Agent prints the preview link in chat
Tap-to-preview: the agent built + served the page and printed `http://localhost:8090 — open preview` as a tappable link in chat (the managed-memory teaching at work); tapping it opens the Preview tab.

Preview tab framing the live site
The Preview tab framing the live `http://127.0.0.1:8090` site — "Hello from AgentNet" rendered in-app.

Drawer after restart
RECENTS [02] in the drawer — both chat sessions persisted across an app restart.

**Video walkthrough (MP4)**

`N/A - Android's screenrecord captured only ~4 frames over 72s because the WebView surface is excluded from screen capture (a platform limitation, not a bug in this PR). The before/after screenshots above ARE the stepped walkthrough; the backend/domain rows below show the same flow end-to-end from the server side.`

**Backend logs**

Observed sequence from the on-device run — the announce path firing end-to-end:

```
[agent]       build complete → serving in background: python3 -m http.server 8090
[chat-server] POST /preview/announce {"port":8090}  (loopback, no Origin → accepted)
[chat-server] broadcast previewStatus {port: 8090, url: "http://127.0.0.1:8090"} → all connected clients

$ curl -i http://127.0.0.1:8090/
HTTP/1.0 200 OK
...
<title>Hello from AgentNet</title>
```

**Frontend logs**

Webview received the `previewStatus` SSE frame (`{port: 8090, url: "http://127.0.0.1:8090"}`) and the Preview tab framed the URL in response — corroborated by the `143-preview-rendered.png` screenshot above showing the framed page.

**Real-LLM trajectory**

N/A - no model or prompt-template change; the new memory block is surface-gated config, and the agent's unprompted serve-and-announce behavior is already evidenced by the backend log sequence above.

**Domain artifacts**

- Generated site: `/root/anet-demo/index.html` — built by the agent, served on port 8090, returns HTTP 200 with `<title>Hello from AgentNet</title>` (see backend logs).
- Persistence: 2 sessions on disk, confirmed via the RPC session list and the RECENTS [02] drawer screenshot after restart.
- Test results: 5/5 new `preview` unit tests pass; `tsc --noEmit` clean (see Code quality above for the pre-existing-on-main vitest failures).

### DIFF
--- a/packages/core/src/memory/previewSection.ts
+++ b/packages/core/src/memory/previewSection.ts
@@ -6,13 +6,11 @@
 //
 // GATED on AGENTNET_PREVIEW_PORT: only the surface that actually HAS the preview tab + announce
 // endpoint (surfaces/localhost, incl. the Android app) sets it. On the CLI/VSCode surfaces the
-// env is unset → the block renders empty and spliceMarkedBlock removes it, so an agent there is
-// never told to hit an endpoint that doesn't exist.
+// env is unset → the block renders empty: an existing block is spliced out, and a file without
+// our markers is left untouched (no write at all), so an agent there is never told to hit an
+// endpoint that doesn't exist.
 
-import { readFile, writeFile } from "node:fs/promises";
-import { join } from "node:path";
-import { claudeMemoryDir, codexAgentsFile } from "../core/paths.js";
-import { spliceMarkedBlock } from "./convert/codex.js";
+import { memoryFileFor, spliceIntoFile } from "./skillsSection.js";
 
 const START = "<!-- agentnet:preview:start -->";
 const END = "<!-- agentnet:preview:end -->";
@@ -29,18 +27,6 @@ function renderBlock(port: string | null): string {
   return `${START}\n${line}\n${END}`;
 }
 
-/** Splice (replace-or-remove) the preview block into a file, preserving everything outside
- *  our markers. Creates the file if missing. */
-async function spliceIntoFile(file: string, block: string): Promise<void> {
-  let existing = "";
-  try {
-    existing = await readFile(file, "utf8");
-  } catch {
-    /* file doesn't exist yet — spliceMarkedBlock returns the block alone */
-  }
-  await writeFile(file, spliceMarkedBlock(existing, block, START, END));
-}
-
 /**
  * Update the "in-app preview" section for one runtime. Renders the hint ONLY when
  * AGENTNET_PREVIEW_PORT is set (the surface with the PREVIEW tab); otherwise removes the block.
@@ -49,9 +35,7 @@ async function spliceIntoFile(file: string, block: string): Promise<void> {
 export async function updatePreviewSection(cli: "claude" | "codex", cwd: string): Promise<void> {
   try {
     const port = process.env.AGENTNET_PREVIEW_PORT || null;
-    const block = renderBlock(port);
-    const file = cli === "claude" ? join(claudeMemoryDir(cwd), "MEMORY.md") : codexAgentsFile(cwd);
-    await spliceIntoFile(file, block);
+    await spliceIntoFile(memoryFileFor(cli, cwd), renderBlock(port), START, END);
   } catch {
     /* never throw from preview-section sync */
   }
--- a/packages/core/src/memory/previewSection.ts
+++ b/packages/core/src/memory/previewSection.ts
@@ -1,0 +1,58 @@
+// "In-app preview" memory section — a managed, auto-updated block that tells the agent the
+// AgentNet app has a PREVIEW tab, so when it builds/runs a web app it should serve it on a
+// loopback port and hand the user the http://localhost:PORT link (which the chat turns into a
+// one-tap preview). Mirrors skillsSection.ts: a SETTER over a marker-delimited block, spliced
+// into claude's MEMORY.md / codex's AGENTS.md, injected every session — no system-prompt nudge.
+//
+// GATED on AGENTNET_PREVIEW_PORT: only the surface that actually HAS the preview tab + announce
+// endpoint (surfaces/localhost, incl. the Android app) sets it. On the CLI/VSCode surfaces the
+// env is unset → the block renders empty and spliceMarkedBlock removes it, so an agent there is
+// never told to hit an endpoint that doesn't exist.
+
+import { readFile, writeFile } from "node:fs/promises";
+import { join } from "node:path";
+import { claudeMemoryDir, codexAgentsFile } from "../core/paths.js";
+import { spliceMarkedBlock } from "./convert/codex.js";
+
+const START = "<!-- agentnet:preview:start -->";
+const END = "<!-- agentnet:preview:end -->";
+
+function renderBlock(port: string | null): string {
+  if (!port) return ""; // no preview surface → remove the block entirely
+  const line = [
+    "In-app preview: this app has a PREVIEW tab that frames a local web server in an iframe.",
+    "When you build or run a web app for the user, start its dev server in the BACKGROUND on a",
+    "loopback port, then give the user the plain http://localhost:PORT link in your reply — it",
+    'becomes a one-tap "open preview" button. (Optional: POST',
+    `http://127.0.0.1:${port}/preview/announce with {"port":N} to open it directly, {"port":null} to clear.)`,
+  ].join("\n");
+  return `${START}\n${line}\n${END}`;
+}
+
+/** Splice (replace-or-remove) the preview block into a file, preserving everything outside
+ *  our markers. Creates the file if missing. */
+async function spliceIntoFile(file: string, block: string): Promise<void> {
+  let existing = "";
+  try {
+    existing = await readFile(file, "utf8");
+  } catch {
+    /* file doesn't exist yet — spliceMarkedBlock returns the block alone */
+  }
+  await writeFile(file, spliceMarkedBlock(existing, block, START, END));
+}
+
+/**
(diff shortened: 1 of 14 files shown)


RemilioNubilio:
@zo-sol — no urgency on this, and happy to close it if it's not a direction you want.

Flagging it now only because of #160: that flow ends with a starter template dropping a ready-to-run prompt into a fresh session, i.e. a new user immediately builds something. This PR is the other half of that moment — a PREVIEW tab that frames a loopback dev server, plus tap-to-preview on the `http://localhost:PORT` links the agent already prints in chat. On a phone especially, "I built a thing" and "I can see the thing" are otherwise separated by an app switch.

Two details that matter if onboarding is being restructured around *connect one engine and start building*:

- It's additive — a 5th dock tab and an announce route; no existing screen changes, nothing removed.
- The agent learns the capability from a managed memory block, so a template-driven build can end in a working preview without the user being told to run anything. That's the part I'd expect to pair well with starter templates.

Built and tested end to end on a Seeker (screenshots in the body). Entirely your call on shape and timing.

zo-sol:
Closing this. We've moved external links to open in the system browser, and a locally served preview on 127.0.0.1:PORT already opens there from the plain link, so framing the same server in-app runs against that direction. It also adds a permanent fifth bottom tab (widening the nav dock to five segments) that shows even when there is nothing to preview, plus an injected agent-memory section and a loopback server to maintain. If we want this later, we can keep just the agent hint to serve builds and hand back a browser link, without a dedicated tab.

RemilioNubilio:
Thanks for the detail, and for the shape you described. We will rebuild it as exactly that: just the agent hint to serve builds and hand back a plain browser link, with no tab, no announce route, and no loopback server to maintain.

Two questions before we start.

Reopen is the one gap the hint alone does not close. Once the user is in the browser and switches back to the app, the way back to the preview is the browser's own tab in recents, or scrolling to the link in the transcript. Is that fine by you, or would you want an affordance that exists only while a dev server is answering and is not there when there is nothing to preview? That shape costs a liveness check we would have to keep honest, so it is your call whether it is worth it. If recents and the transcript are fine, say so and we will build it that way.

On the link itself, I may be reading this wrong. The comment at :165 states the rule as our app UI stays in the WebView and any other http(s) link opens in the external browser, but `openExternally` matches on host alone (MainActivity.kt:286-288), so a dev server on another loopback port reads as our own UI on `Paths.PORT`. `shouldOverrideUrlLoading` (:170) then keeps it in place, and `onCreateWindow` (:183) routes through the same check, so a `target=_blank` link does not escape either. With no `onBackPressed`, `OnBackPressedCallback` or `goBack` anywhere under surfaces/android, back finishes the activity rather than returning to our UI, which is the same failure the comment at :168 calls out for claude.ai. Is in-place navigation intended for foreign loopback ports, or would a port check be welcome so they take the same ACTION_VIEW path as any other external host?

Issue #101, opened 2026-07-02
Study iq-chan: table of blockchain-SNS patterns AgentNet can borrow


zo-sol:
## Why

AgentNet already ships SNS-style features (agent profiles, blog posts, comments) backed by on-chain storage. We have built a decentralized social app on Solana before, so instead of reinventing, we should mine our own prior art and borrow what already works.

**Primary repo to study: [IQCoreTeam/iq-chan](https://github.com/IQCoreTeam/iq-chan)** (BlockChan) - a fully decentralized imageboard on Solana. Threads, replies, boards, permanent on-chain posts, and a read-only gateway cache. This is rich prior art for exactly the surfaces AgentNet has today: **threading / replies, posting UX, board & profile structure, and on-chain post storage.**

## Task

1. **Pull iq-chan locally.**
2. **Have an agent read every file** (not just the README) - code, SDK/contract usage, and UI flows.
3. Produce a **learning table** of everything AgentNet can borrow.

## Deliverable - the table

| Idea / Feature | How iq-chan does it | What AgentNet can borrow | Effort / priority |
|---|---|---|---|

Seed ideas to look for (extend as you read):
- **Reply / threading** - nested replies on blog posts and comments (AgentNet is flat today)
- **Profile setup** - how identity/profile is created, named, and edited
- **On-chain post/thread storage** - schema, chunking/paging, cost, and the cache/gateway pattern
- **Posting UX** - compose, preview, image handling, confirmation
- **Read-only gateway cache** - serving on-chain data cheaply without hammering RPC
- **Moderation / spam handling** on a permissionless chain

## Secondary reference (optional)

If useful while thinking about search and feed-building, also glance at **[IQCoreTeam/iq-wide-web](https://github.com/IQCoreTeam/iq-wide-web)** for total information retrieval / indexing ideas (naming, linking, discovery over on-chain content). Not the focus of this issue - just a reference to pull from if an indexing question comes up.

## Notes

- This is a **research / synthesis** issue: the output is the table plus short recommendations, not code.
- When we later act on it, keep our standing constraint in mind - do not multiply RPC/API calls. The gateway-cache pattern in iq-chan is exactly where to learn the efficient approach.

mega123-art:
## Study: what AgentNet can borrow from iq-chan (BlockChan)

Read all of iq-chan (code, hooks, SDK usage, scripts) + glanced at iq-wide-web.

### Architecture (what iq-chan actually is)

[...]

Issue #93, opened 2026-07-01
Port mobile marketplace features to CLI surface (full parity: 16 render + 4 form gaps)


mega123-art:
## Summary

Port the mobile (webview) marketplace features to the CLI surface. Original scope said "just the agent list" — a full CLI-vs-webview sweep shows the gap is much larger: **16 render-only gaps + 4 data/form gaps**. Most are render-only because the CLI already fetches the data via `MarketApi` / `marketplaceEnv`.

## Investigation findings

Surfaces live in `surfaces/{android,cli,vscode,webview}`.

- **Mobile app == the React webview** (`surfaces/webview/src/market`) — source of truth. Full marketplace UI.
- **VSCode surface** loads the **same React webview** (`chatHtml`) → already at full parity. **No work needed.**
- **CLI surface** (Ink UI, `surfaces/cli/src/views/SkillMarket.tsx`) is the only gap. It already fetches the data (`stars`, `verifiedRepos`, notes, etc. flow through `marketplaceEnv` / `MarketApi`) but the render layer ignores most of it.

## Gap table

| Feature | Webview | CLI | Data available to CLI? | Gap type |
|---|---|---|---|---|
| Agent directory | Sortable list, sticky self, wallet search | 12-item list, arrow nav, no search | `listAgents()` | Render (search + sort) |
| Agent stars/tier | Tier tag + gauge + ladder | Not shown | `Reputation.stars` | Render |
| Agent earned (SOL) | On card | Not shown | `Reputation.totalEarned` | Render |
| Verified GitHub repos | Folder cards, star gauge, "show all" modal | Not shown | `AgentProfile.verifiedRepos` | Render (whole feature) |
| Blog posts | Carousel: title, image, text, date, link | Not shown | `AgentProfile.notes` | Render (whole feature) |
| Agent comments | Full stack w/ avatar, date, link | First 4 truncated | `AgentProfile.notes` | Render (full view) |
| Blog composer (self) | FAB → write blog | Not shown | `postAgentNote()` | Render |
| Required skills | Checkmarks + prices + "Collect all" | Plain name list | `SkillDetail.requiredCards` | Render |
| SKILL.md | Full text | Truncated 1200 chars | `SkillDetail.skillText` | Render |
| Skill comments | Full stack | First 4 truncated | `SkillDetail.notes` | Render (full view) |
[...]

PR #157, opened 2026-08-14
Core/API glue the CLI design-parity pass needs (parentId, buy error code, airdrop, GitHub trio)

RemilioNubilio:
# The CLI parity plan's four core/API glue items, shipped separately so the design pass stays purely visual

`plans/cli-design-parity.md` lists four items under **"### Core/API glue needed (not just UI)"** (lines 111–117), each blocking a UI tab of the parity redesign. This branch ships exactly those four — one commit each, in the plan's order — plus a fifth commit that pins them at compile time, so the UI pass never dead-ends on a missing core surface. Built to unblock the parity pass — happy to reshape any of it to whatever the UI work wants. No UI, rendering, or layout is touched: the entire CLI diff is interface-declaration lines and type-level pins in `SkillMarket.tsx`.

- **Line 113** — *"`MarketApi.postAgentNote` drop of `parentId` -> add it (core supports)"* [tab 22, reply composer]: `parentId` added as the optional 6th parameter. Core already carries it end-to-end (GH #101 reply threading — `marketplaceEnv` → notes layer); the CLI type was the drop point, so the tab-22 composer could never have passed it.
- **Line 114** — *"`MarketApi.buySkill` drops `code` -> add, detect `insufficient_funds`"* [tabs 26/08/24]: the result type gains `code?: "insufficient_funds"`. The `{code}` field is how the classification already travels through `marketplaceEnv`/`buyResult`; the CLI type dropped it, so a FUNDING panel had nothing to key off.
- **Line 115** — *"Wire `marketplaceEnv.airdrop()` (devnet)"* [tab 26]: `airdrop()` declared as a **required** `MarketApi` member (the CLI's api object *is* `marketplaceEnv`, so the requirement is free), and the missing delegation added in `surfaces/vscode/src/extension.ts` — which un-breaks the webview FundModal's airdrop request that previously hit the dispatcher's missing-capability guard and silently did nothing. The devnet-only guard lives in core.
- **Lines 116–117** — *"GitHub `submitGithubToken`/`getGithubStatus`/`registerWorkRepo`: message types exist but `marketplaceEnv` exposes none -> needs CORE additions"* [tabs 24/26]: the trio becomes `marketplaceEnv` members (`packages/core/src/skill-market/ingest/env.ts`), deferring to the `rpc.ts` token store (0600 file, only the masked tail leaves) and `verifiedWork.ts`. Two contracts hardened while consolidating: `registerWorkRepo` keeps the *"Connect a wallet first."* refusal both surface copies guarded with — `registerVerifiedWork` needs no signature, just the address string, so an env built before a wallet is connected would otherwise commit whatever address it holds as the public `.agentnet` marker and index it as the user's identity; and `submitGithubToken` answers a failed save (unwritable tokens dir, full disk) as `{ hasToken: false, error }` instead of rejecting — the dispatcher spreads the result straight into `githubStatus` with no catch of its own, so a rejection was an unhandled rejection in the host and a modal hanging on "Saving…" forever. `githubStatus` and the `ChatEnv` signature gain the optional `error` field and the webview token form renders it. vscode's hand-rolled copies become 3-line delegations (~20 duplicated lines deleted); localhost keeps its own handlers on purpose (pre-wallet token save + `clearGithubToken`, which has no env member yet) — the env doc comment states exactly that split rather than overclaiming.
- **Pins** — `SkillMarket.tsx` gains three zero-runtime type-level asserts (`Parameters[...][5]` for `parentId`, the `["code"]` index for `insufficient_funds`, an extends-check for `airdrop`). The env→`MarketApi` assignment only checks one direction — dropping an optional widening would still typecheck until a consumer calls it — so the pins make a revert fail `tsc` *here*, today.
- `packages/core/src/skill-market/ingest/env.spec.ts` grows 12 specs beside its existing publish suite: `parentId` forwarding ×2, buy-code classification ×2, airdrop devnet guard ×1, github trio ×7 (status without a token, store + refreshed mask, a failed save answered as an error, the no-wallet and no-token refusals, the registration call shape, error passthrough).

Honest scope note: three of the four plan items were already true inside core — this branch makes the API surface *admit* it (widenings + specs + pins). The behavioral changes live in the airdrop delegation (was a silent no-op in vscode), the github-trio consolidation, and the two hardened trio contracts above (wallet-gated registration, save failures answered instead of thrown).

Verified: `tsc --noEmit` clean for core, cli, and vscode (cli and vscode confirmed clean pre-change too, so the widenings introduced no looseness); `env.spec.ts` passes 15/15 under `AGENTNET_NETWORK=devnet`; all four surface builds pass.

## Relates to

`plans/cli-design-parity.md` §"Core/API glue needed (not just UI)", lines 113–117, serving tabs 22 (Agent Comments reply composer), 24 (Skill Detail `[G] REGISTER WORK`), 26 (Connections + Funding / FUNDING panel), and the tab-08 buy path. Also GH #101 (reply threading) and issue #93 (verified work). This is deliberately *only* the glue: the call sites that branch on `code`, the FUNDING panel, and the reply composer are the parity pass's UI work — built to unblock that pass, and happy to adjust any of these shapes to whatever the UI work wants.

## Risks

**Low.** The widenings are additive-optional — I traced every caller: the CLI's `doBuy`/`doCollectAll` read only `ok`/`error`, the dispatcher's spread targets a union that already carries `code` (and now `error`), and the vscode delegations are covariant-safe. The vscode airdrop delegation activates an existing core path with its own devnet guard. The github delegation drops vscode's local wallet check because the refusal now lives in the env member itself — enforced in one place for every surface that delegates, and spec-pinned. localhost intentionally keeps its own github handlers — its env omits the trio, so the dispatcher's capability guards fall through to localhost's guest-aware handlers exactly as before.

## Background — what & why

The parity plan's biggest tabs each dead-end on a missing core surface: a reply composer with no `parentId` to send, a FUNDING panel with no `insufficient_funds` signal and no `airdrop()` to offer, a REGISTER WORK flow whose message types exist but whose `marketplaceEnv` exposes nothing. Landing the glue first, in its own reviewable branch, means the design pass stays purely visual — and the pins mean the glue can't silently rot while it waits to be consumed.

Start at the four new describes in `packages/core/src/skill-market/ingest/env.spec.ts` — they pin the behavior the widened types now advertise (parentId reaching the notes layer, the broke-wallet classification, the mainnet airdrop refusal, the token-store round-trip, the no-wallet and no-token refusals, and the failed save answered as an error rather than thrown).

1. `cd packages/core`
2. `AGENTNET_NETWORK=devnet npx vitest run src/skill-market/ingest/env.spec.ts` → 15/15 pass
3. `npx tsc --noEmit` → clean (repeat in `surfaces/cli` and `surfaces/vscode`)
4. Mutation check: revert any one widening in `SkillMarket.tsx`'s `MarketApi` — `tsc` fails with exactly one error anchored at that widening's pin.

### Code quality

Held to five rules — each verified against this diff, not asserted:

1. **No meaningless wrappers with identical input/output** — ✅ the new env members own real logic: the wallet refusal that keeps an unconnected env from committing an identity marker, the token-store round-trip returning the refreshed masked status (one render path for both submit and status), the failed-save-as-answer contract, the no-token refusal, the indexer call, error mapping. The vscode side is delegation at the host boundary — the established pattern of every neighboring market field — and each delegation *replaced* a full duplicate implementation (github trio) or a silent gap (airdrop); none renames a call.
2. **Scan existing functions first and reuse; never two functions with the same purpose** — ✅ this branch is that rule applied: three near-identical github flows collapse toward one core implementation; `airdrop` reuses the existing env member and its devnet guard rather than growing a second faucet path; the specs extend the existing `env.spec.ts` file and its `rpc.js` mock factory instead of a new harness.
3. **No type variables that won't be reused; inline them** — ✅ one local `AssertTrue` helper serves all three pins and is erased at compile; everything else is inline widening of existing signatures (`error?: string` rides the existing `githubStatus` union member and `ChatEnv` method type) — no new exported types anywhere.
4. **No non-essential parameters** — ✅ `parentId` is the single field reply threading needs and core already serializes; `code` appears only when the classification fires and is typed to the one value the panel keys off; `airdrop()` takes nothing and returns exactly what the FundModal renders; `error` appears only on a failed save and the token form renders it.
5. **Keep responsibilities separated; if the design feels wrong, say so** — ✅ `env.ts` stays the single market-capability surface; surfaces stay transport; the trio's guards live in the env member, not per-surface. Two things said out loud: localhost deliberately keeps its own github handlers (its pre-wallet token save and `clearGithubToken` have no env equivalent yet — a future "disconnect GitHub" wants a small env addition, cheap follow-up; the env doc comment records this split so nobody reads it as an oversight); and until tabs 22–26 land, the trio's only consumer is the dispatcher — that is the point of a glue branch, but naming it so nobody mistakes the coverage.

`tsc --noEmit` clean (core · cli · vscode) · `env.spec.ts` 15/15 (12 new) · pins mutation-verified · no regressions — the failing vitest specs under a default env (rpc/seed/dasSource/skillSource + two stale session slash-command specs) reproduce identically without this branch's changes: pre-existing on `main` (d6a0881) and env-dependent (`seed.ts`'s mainnet flip vs stale devnet specs), zero failures attributable to this branch. Based on latest `main`, merges clean, independent of the other branches.

## Evidence

| Row | Artifact |
|---|---|
| Before/After screenshots | N/A - no UI change to show: the CLI diff is type declarations + compile-time pins, the vscode diff is host-side handler wiring behind existing UI (the one webview line renders an error string into an existing error slot) |
| Video walkthrough (MP4) | N/A - no user-visible flow changes; the one behavioral fix (FundModal airdrop no longer a silent no-op) is a request that previously died in the dispatcher, evidenced by the delegation + core's devnet-guard spec |
| Backend logs | `env.spec.ts` run — 15/15: parentId forwarded to the notes layer (reply + top-level) ×2, `insufficient_funds` classification (broke wallet vs other failure) ×2, airdrop refused off-devnet ×1, github trio (status masked-tail, token save round-trip, failed save answered as `{ hasToken:false, error }`, no-wallet refusal, no-token refusal, register success, register error) ×7, plus the 3 pre-existing publish specs |
| Frontend logs | N/A - the only client-side change is one line rendering `githubStatus.error` in the existing token form; the contract beneath it is spec-pinned at the env boundary |
| Real-LLM trajectory | N/A - no agent, model, or prompt changes; API-surface glue for the parity plan |
| Domain artifacts | The compile-time pins are the artifact: reverting any one widening on this branch yields exactly one `tsc` error anchored at its pin (mutation-verified per-widening), so the plan's glue cannot silently regress before the UI pass consumes it |

### DIFF
--- a/packages/core/src/chat/marketMessages.ts
+++ b/packages/core/src/chat/marketMessages.ts
@@ -150,7 +150,7 @@ export type MarketEvent =
   | { type: "balance"; lamports: number | null } // wallet SOL balance (null = read failed)
   | { type: "airdropResult"; ok: boolean; lamports?: number; error?: string } // devnet faucet result (lamports = the new balance)
   // GitHub verified-work registration results (issue #93 parity)
-  | { type: "githubStatus"; hasToken: boolean; masked?: string }
+  | { type: "githubStatus"; hasToken: boolean; masked?: string; error?: string } // error = token save failed
   | { type: "workRepoRegistered"; ok: boolean; count?: number; repo?: string; error?: string }
   | { type: "skillActive"; name: string; origin: "nft"; mint: string } // nft skill fired -> casting cue
   | { type: "rpcStatus"; status: RpcStatus } // DAS-ready? which source? (issue #23)
--- a/packages/core/src/chat/session.ts
+++ b/packages/core/src/chat/session.ts
@@ -92,7 +92,7 @@ export interface ChatEnv {
   // owner/name repo against the given skill mints. All three defer to core (rpc.ts token
   // store + verifiedWork.ts). A surface that can't do this simply omits them.
   getGithubStatus?(): Promise<{ hasToken: boolean; masked?: string }>;
-  submitGithubToken?(token: string): Promise<{ hasToken: boolean; masked?: string }>;
+  submitGithubToken?(token: string): Promise<{ hasToken: boolean; masked?: string; error?: string }>; // error = save failed (answered, never thrown)
   registerWorkRepo?(repo: string, skillMints: string[]): Promise<{ ok: boolean; count?: number; repo?: string; error?: string }>;
   // make-skill: publish a new skill from the UI. priceSol is the human SOL string; the
   // host converts to lamports and calls core publishSkill. Returns the new mint on success.
--- a/packages/core/src/chat/ui/webview.ts
+++ b/packages/core/src/chat/ui/webview.ts
@@ -4215,6 +4215,9 @@ export function chatHtml(): string {
       link.href = 'https://github.com/settings/tokens/new?scopes=repo&description=AgentNet';
       link.target = '_blank'; link.rel = 'noopener noreferrer'; link.textContent = 'Create a token ↗';
       const err = document.createElement('div'); err.className = 'rr-err';
+      // a failed save answers githubStatus { hasToken:false, error } — show it here so the
+      // re-rendered form explains the failure instead of looking like a silent reset.
+      if (status && status.error) { err.textContent = status.error; err.style.display = ''; }
       const btn = document.createElement('button'); btn.className = 'rr-btn'; btn.textContent = 'Save token';
       btn.addEventListener('click', () => {
         const t = inp.value.trim(); if (!t) return;
--- a/packages/core/src/skill-market/ingest/env.spec.ts
+++ b/packages/core/src/skill-market/ingest/env.spec.ts
@@ -3,6 +3,7 @@ import { marketplaceEnv } from "./env.js";
 import { publishSkill as corePublishSkill } from "../../nft/skill.js";
 import { publishWorkflow as corePublishWorkflow } from "../../nft/workflow.js";
 import { postAgentNote as corePostAgentNote } from "../../notes/notes.js";
+import { getNetwork } from "../../core/seed.js";
 
 vi.mock("../../nft/skill.js", () => ({
   publishSkill: vi.fn().mockResolvedValue("mockSkillMint"),
@@ -30,6 +31,13 @@ vi.mock("./index.js", () => ({
 vi.mock("../../core/rpc.js", () => ({
   resolveRpcUrl: vi.fn().mockResolvedValue("http://localhost:8899"),
 }));
+
+// Partial mock: getNetwork becomes controllable (the airdrop guard needs a non-devnet
+// answer) while keeping its real devnet default and the rest of the seed module real.
+vi.mock("../../core/seed.js", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("../../core/seed.js")>();
+  return { ...actual, getNetwork: vi.fn(() => "devnet" as const) };
+});
 
 // Partial mock: only the agent-note write + re-read are stubbed; the rest of the
 // notes module stays real so nothing else in env's import graph changes shape.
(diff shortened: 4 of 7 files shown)


zo-sol:
Closing this. Three of the four pieces are already on main from parallel work: parentId note threading (notes/notes.ts), the insufficient_funds buy code (skill-market/ingest/env.ts, chat/session.ts), and the devnet airdrop helper (skill-market/ingest/env.ts). The only net-new part is bringing GitHub verified-work to the CLI (getGithubStatus / submitGithubToken / registerWorkRepo plus the CLI wiring); the app and VS Code already ship this flow. That belongs with the CLI first-class-surface work in #167, so tracking it there instead of merging this mostly-superseded branch. The GitHub piece can be reimplemented or cherry-picked under #167 if we pick it up.

RemilioNubilio:
Understood, and agreed the three overlapping pieces were already covered by the parallel work.

We can take the GitHub verified-work part to #167 (getGithubStatus / submitGithubToken / registerWorkRepo plus the CLI wiring) if you want it from us rather than keeping it in-house. Just say the word and we will pick it up there.

PR #146, opened 2026-08-13
Session sync's missing half: the mysessions on-chain index — opt-in cross-device session discovery

RemilioNubilio:
# Session sync's missing half: the `mysessions` on-chain index — opt-in cross-device session discovery

Implements the unbuilt piece of `plans/offchain-session-sync.md` §4–5. The off-chain half already ships (encrypted paged blobs over `StorageAdapter`, wallet-derived X25519 keys, cross-device resume) — but a brand-new device still has no way to *learn what sessions its wallet owns* until the right cloud storage is configured. The plan's answer is the `mysessions` table, and its scaffolding was already committed waiting: `mysessionsHint()` in `core/seed.ts` (zero callers until now), the `Session` row type in `core/types.ts`, and `ensureTable`/`writeRow` with `writers` support in `core/chain.ts`. This PR wires them together, following the `notes.ts` table recipe.

## What it does

- **One row per NEW sessionId, written once at creation** — exactly the plan's §5.3 economy: updates never touch the chain (integrity lives in the blob's authenticated encryption, not a hash). The write hooks the canonical-id reveal in `runtime/index.ts` (`onSessionId` for fresh sessions, `forkSession` for forks — the two places a canonical id is minted). Resumes and ephemeral sessions never index.
- **Titles stay off-chain.** The plan marks `title` "omit if sensitive"; this ships that as the *default* — a table row is public forever, and titles already live inside the encrypted blobs. The column is declared (columns are fixed at table creation, `SESSION_COLUMNS` mirrors `REVIEW_COLUMNS`' `meta` escape hatch) but the writer never fills it.
- **Opt-in per wallet, default OFF.** A row write is a real Solana transaction — fees, and a signature prompt on web wallets. Consent lives in `config.json` as an address-keyed map, not one device-global flag: a CLI keypair opting in must not surprise-prompt a Phantom wallet sharing the same `AGENTNET_HOME`.
- **Best-effort everywhere.** Chain trouble (RPC down, the localhost guest wallet's throwing signer, a squatted table) resolves to a warn and the chat continues untouched — the index is a discovery layer, never a dependency.
- **Paid writes never dedup against a partial view.** The backfill (`/sessionsync on` publishes pre-existing sessions) dedups against the chain *and* a local `chain-index/{wallet}.json` cache (the gateway's view can lag its own write-notify), refuses a truncated read window, rejects concurrent re-entry, and persists paid-for ids even when a mid-run write throws. Every one of those guards exists because writing a duplicate row costs the user real money.
[...]

PR #158, opened 2026-08-14
Bring the session index to the webview and vscode surfaces (completes #146's open decision)

RemilioNubilio:
# The on-chain session index reaches the webview and vscode surfaces, so cross-device discovery isn't CLI-only

#146 shipped the `mysessions` on-chain index (plans/offchain-session-sync.md §4–5) with a CLI surface (`/sessionsync`) and deliberately left one decision open: *"whether the webview drawer should grow the same toggle once the CLI shape settles."* The shape settled — this PR answers that decision. The webview (browser/Android) and vscode surfaces get the same opt-in, the same consent framing, and the same status line, all through **one dispatcher round-trip** so the index policy stays in core with zero duplication. And it matters most exactly here: a phone is the archetypal "second device" the index exists for. No CLI view/layout files are touched — the CLI screens belong to the design-parity pass.

- `surfaces/webview/src/transport/protocol.ts` — two new union members: a `sessionIndex` request (`status | on | off | backfill`) and its `sessionIndexStatus` result (`enabled`, plus optional `listed` / `elsewhere` / `truncated` / `written` / `error`), each field documented at the declaration.
- `packages/core/src/chat/session.ts` — one dispatcher case beside the cloud actions, calling core's `setSessionIndex` / `backfillSessionIndex` / `sessionIndexStatus` directly. Hosts supply only the new optional `ChatEnv.signingWallet()` — the wallet that can *sign transactions*, distinct from `walletAddress()` (display): the localhost guest key signs messages, never transactions, so it returns null and the paid toggle can never reach an unsignable wallet. `on` runs the one-shot backfill (CLI parity: the list starts complete). `off` is the safety exit — config-only, confirms immediately, no chain read, so a dead RPC can never paint an error over an action that fully succeeded. A failed action reports its reason with the toggle state **re-read from config**, so the switch stays truthful. The chain work runs *off* the strictly-ordered pump (fire-and-forget, CLI parity again): `on`/`backfill` mean one real Solana transaction per unindexed session — behind a UI signing prompt per row on localhost — and awaiting that in the dispatcher would freeze every queued message (chat sends, stops, session opens) behind wallet prompts. Backfill re-entry is rejected in core, so overlapping requests fail loudly instead of double-paying.
- `surfaces/webview/src/chat/Sessions.tsx` + `store.tsx` — a wallet-gated `ProgressiveMenuRow` ("Session sync · On-chain list") in the drawer settings, matching the Storage row idiom, opening a sub-screen: toggle carrying the real-transaction consent copy (fees + signing prompts, id-only — never content, and that turning on lists existing sessions right away), one-shot "Backfill older sessions", and the "N listed · M on other devices" status line. Store gains `state.sessionIndex` + its reducer case; controls hold (`syncBusy`) until the dispatcher's echo lands.
- `packages/core/src/chat/ui/webview.ts` (the vscode-hosted webview) — a minimal "Session sync" section in the wallet menu: status line, lowercase turn on/off + backfill links, and a consent hint shown until opted in that carries the *full* disclosure the other two surfaces give — per-listing tx cost, id-only, and "turning on lists your existing sessions right away" — so `on`'s immediate one-shot backfill never surprises anyone on any surface. Hydrated on wallet-menu open (one line added to the existing `getBalance`-on-open hook), never polled — a status read hits the chain.
- Hosts: `surfaces/vscode/src/extension.ts` returns the local keypair (signs in-process — listing costs only the tx fee); `surfaces/localhost/src/index.ts` returns the *connected* wallet only, never the guest.
- `packages/core/src/chat/session.spec.ts` — a 7-test "on-chain session index" describe on the file's existing mock-transport harness: absent hook / null wallet stay silent, `status` never writes a paid row, `on` flips config and runs the one-shot backfill, `off` confirms from config with the chain dark, a failed action echoes the config-read toggle, and a hung RPC never blocks later dispatch. Also repairs the file's two stale slash-command specs (the sessions feed grew `running`, ownedSkills grew `meta` + lazy-import latency) so its suite runs green.

Verified: `tsc --noEmit` clean for core, localhost, and vscode; `session.spec.ts` + `sessionIndex.spec.ts` pass 44/44 and the full core suite passes 277/277 under `AGENTNET_NETWORK=devnet` (the specs' pinned default); all four surface builds pass (webview vite, localhost/vscode/cli tsup).

## Relates to

#146 — completes the open decision it flagged ("whether the webview drawer should grow the same toggle once the CLI shape settles"); the branch carries #146's commit beneath this one, so this PR shows two commits until #146 lands and then collapses to just the surface work (rebased or not — it merges clean either way, verified against `6ab2fb2`). Happy to hold it until #146 is decided so you're only reading one diff. Implements the cross-surface half of `plans/offchain-session-sync.md` §4–5. Deliberately stays out of the CLI screens the design-parity plan is reworking.

## Risks

**Low.** Additive protocol messages; the dispatcher case is inert unless a host supplies `signingWallet` *and* the UI sends the message. The paid-write machinery itself is #146's core layer, unchanged here — this case never writes on `status`, never touches the chain on `off`, and inherits core's re-entry/truncation/dedup guards on `on`/`backfill`. The one behavior worth naming: chain work is fire-and-forget off the dispatcher pump precisely so a slow RPC or a stack of signing prompts can't stall chat dispatch — the spec pins that with a hung-RPC test.

## Background — what & why

The index exists so a brand-new device can *learn what sessions its wallet owns* before any cloud storage is configured — and after #146, only the CLI could opt in or see "N on other devices." The devices where that discovery actually matters — a phone, a fresh vscode install — had no surface for it. One dispatcher case gives all webview-transport surfaces the identical contract without a second copy of the index policy.

Start at the "on-chain session index" describe in `packages/core/src/chat/session.spec.ts` — it drives the real dispatcher over the mock transport with the core index API mocked, so every consent/paid-write contract is pinned at the message boundary.

1. `cd packages/core`
2. `AGENTNET_NETWORK=devnet npx vitest run src/chat/session.spec.ts src/account/sessionIndex.spec.ts` → 44/44 pass (full core suite: 277/277)
3. `npx tsc --noEmit` → clean (repeat in `surfaces/localhost`, `surfaces/vscode`)
4. Mutation check: reverting the `session.ts` case fails the two behavior-change tests; the other five guard the contracts (guest silence, status-never-writes, off-with-chain-dark, truthful error echo, hung-RPC non-blocking).
5. Consent-copy check: all three surfaces disclose the same three facts before opt-in — per-listing tx cost, id-only (never content), and that turning on backfills existing sessions immediately (React sub-screen copy, vscode `ssyncHint`, CLI `/sessionsync`).

### Code quality

Held to five rules — each verified against this diff, not asserted:

1. **No meaningless wrappers with identical input/output** — ✅ `signingWallet()` is not `walletAddress()` renamed: one is a display string, the other the transaction-capable `Wallet` or null — on localhost they *disagree* (a guest has an address but returns null here), and that disagreement is the gate. The dispatcher case calls the core index API directly; no forwarding layer was added anywhere.
2. **Scan existing functions first and reuse; never two functions with the same purpose** — ✅ the case reuses #146's core API unchanged (`setSessionIndex` / `backfillSessionIndex` / `sessionIndexStatus`); the React screen reuses `ProgressiveMenuRow` / `Toggle` / `SettingsSubHeader`; the vscode menu reuses the `wmSection` idiom and adds one line to the existing menu-open hook; the tests join the file's existing mock-transport harness. The index policy exists in exactly one place.
3. **No type variables that won't be reused; inline them** — ✅ both protocol messages are inline union members; `store.tsx`'s `sessionIndex` state is typed inline; the only type import added to `session.ts` is the existing `Wallet`.
4. **No non-essential parameters** — ✅ the request carries only `action`; every optional response field renders on at least one surface (`listed`/`elsewhere`/`truncated` on both, `written` as the backfill acknowledgment, `error` alongside the config-read `enabled` so the switch stays truthful). `signingWallet` is optional so hosts without the capability change nothing.
5. **Keep responsibilities separated; if the design feels wrong, say so** — ✅ index policy in `core/account`, the round-trip in the dispatcher, hosts supplying one capability, UIs rendering one message. Said-so: a `sessionIndex` request arriving with *no* signing wallet is silently unanswered. Safe today — I verified both UIs wallet-gate the row (the React row via `ProgressiveMenuRow`, and vscode chat panels only exist while a wallet is set) — but the invariant lives in UI gating; if a future surface un-gates it, the protocol should answer with a self-describing error instead. Flagging it rather than pretending the protocol is self-defending.

`tsc --noEmit` clean (core, localhost, vscode — the webview package's 8 TS6133 are byte-identical pre-existing on `main`; its vite build is the gate and passes) · session + sessionIndex specs 44/44, full core suite 277/277 · no regressions — the rpc/seed/dasSource/skillSource vitest failures under a default env are pre-existing on `main` and env-dependent (green under `AGENTNET_NETWORK=devnet`; without it, `seed.ts`'s mainnet flip vs stale devnet specs). Based on `main` (d6a0881) + #146's commit; merges clean.

## Evidence

| Row | Artifact |
|---|---|
| Before/After screenshots | 📎 *(attached on this PR)* webview drawer: locked "Session sync" row for guests → unlocked row → sub-screen with consent copy, toggle off→on, and the "N listed · M on other devices" line; vscode wallet menu: "Session sync" section hint (with the backfill disclosure) → status after turn-on |
| Video walkthrough (MP4) | N/A - two short click flows fully shown by the screenshots; the behavior beneath each click is spec-pinned at the message boundary |
| Backend logs | `session.spec.ts` "on-chain session index" describe 7/7 (guest silence, status-never-writes, on = consent + one-shot backfill, off-with-chain-dark, truthful error echo, hung-RPC non-blocking) + `sessionIndex.spec.ts` 14/14; full core suite 277/277 under `AGENTNET_NETWORK=devnet` |
| Frontend logs | N/A - both UIs are a pure render of the single `sessionIndexStatus` message; the only client state is `store.tsx`'s reducer case, driven end-to-end by the dispatcher spec at the transport boundary |
| Real-LLM trajectory | N/A - no agent, model, or prompt changes; protocol + dispatcher + surface wiring |
| Domain artifacts | The paid-write safety contracts are executable: `status` never writes a row, `off` completes with the chain dark, overlapping backfills are rejected in core, and a guest can never trigger a transaction. The consent copy is uniform across all three surfaces, including the immediate-backfill disclosure. The on-chain layer beneath is #146's live devnet + mainnet proofs (`writers=[owner]` enforcement, real row writes) — untouched by this PR |

### DIFF
--- a/packages/core/src/account/login.ts
+++ b/packages/core/src/account/login.ts
@@ -75,6 +75,30 @@ export async function setSkillShopping(on: boolean): Promise<void> {
   await writeFile(configFile(), JSON.stringify({ ...prev, skillShopping: on }, null, 2));
 }
 
+// On-chain session index toggle (plans/offchain-session-sync.md §4-5). Same raw-key
+// pattern as skillShopping, but the default is OFF and consent is PER WALLET (a map
+// of address → true): indexing a session writes a real Solana transaction — fees,
+// and a signature prompt on web wallets — so a CLI keypair opting in must not drag
+// a Phantom wallet sharing this device's config into surprise prompts. Consumed by
+// account/sessionIndex.ts (every write path gates on it) and the surfaces' toggles.
+export async function getSessionIndex(wallet: string): Promise<boolean> {
+  const raw = await readRawConfig();
+  const map = raw.sessionIndex;
+  return typeof map === "object" && map !== null && (map as Record<string, unknown>)[wallet] === true;
+}
+
+export async function setSessionIndex(wallet: string, on: boolean): Promise<void> {
+  await ensureDir(rootDir());
+  const prev = await readRawConfig();
+  const map = typeof prev.sessionIndex === "object" && prev.sessionIndex !== null
+    ? (prev.sessionIndex as Record<string, unknown>)
+    : {};
+  await writeFile(
+    configFile(),
+    JSON.stringify({ ...prev, sessionIndex: { ...map, [wallet]: on } }, null, 2),
+  );
+}
+
 // First-run setup: record the chosen backend and (for gdrive) run Google sign-in.
 // `openBrowser` is injected by the surface (only used for gdrive).
 export async function initialize(
--- a/packages/core/src/account/sessionIndex.spec.ts
+++ b/packages/core/src/account/sessionIndex.spec.ts
@@ -1,0 +1,204 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { mkdtempSync, rmSync, readFileSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+import { ensureTable, writeRow, readRows } from "../core/chain.js";
+import { SESSION_COLUMNS } from "../core/seed.js";
+import { setSessionIndex, getSessionIndex } from "./login.js";
+import {
+  indexNewSession,
+  listChainSessions,
+  sessionIndexStatus,
+  backfillSessionIndex,
+} from "./sessionIndex.js";
+import type { Wallet } from "../runtime/contract.js";
+
+// The chain layer is fully mocked: these tests pin the GATING (opt-in, dedup,
+// best-effort, truncation refusal) and the row/table shapes handed to it — the
+// layer chain.spec.ts already covers. resolveRpcUrl is mocked so the lazy init
+// never probes a key.
+vi.mock("../core/chain.js", () => ({
+  init: vi.fn(),
+  ensureTable: vi.fn().mockResolvedValue(null),
+  writeRow: vi.fn().mockResolvedValue("txSig"),
+  readRows: vi.fn().mockResolvedValue([]),
+}));
+vi.mock("../core/rpc.js", () => ({
+  resolveRpcUrl: vi.fn().mockResolvedValue("https://api.devnet.solana.com"),
+}));
+
+const wallet = { address: "walletA" } as unknown as Wallet;
+let home: string;
+
+beforeEach(() => {
+  home = mkdtempSync(join(tmpdir(), "agentnet-test-"));
+  process.env.AGENTNET_HOME = home;
+  vi.mocked(ensureTable).mockClear().mockResolvedValue(null);
+  vi.mocked(writeRow).mockClear().mockResolvedValue("txSig");
+  vi.mocked(readRows).mockClear().mockResolvedValue([]);
+});
+
+afterEach(() => {
+  delete process.env.AGENTNET_HOME;
+  rmSync(home, { recursive: true, force: true });
+});
+
+describe("sessionIndex toggle", () => {
+  it("defaults OFF — absent flag is not consent to pay for transactions", async () => {
+    expect(await getSessionIndex("walletA")).toBe(false);
+  });
+
+  it("consent is PER WALLET — one wallet's opt-in never covers another", async () => {
+    await setSessionIndex("walletA", true);
+    expect(await getSessionIndex("walletA")).toBe(true);
+    expect(await getSessionIndex("walletB")).toBe(false);
+    await setSessionIndex("walletA", false);
+    expect(await getSessionIndex("walletA")).toBe(false);
+  });
+});
+
+describe("indexNewSession", () => {
+  it("does nothing while the wallet's toggle is off", async () => {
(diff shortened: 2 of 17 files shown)


zo-sol:
Closing alongside #146, which this builds on. With the on-chain session index deferred (see #146), bringing it to the webview and VS Code surfaces has nothing to stand on. Will revisit together with #146 if we pick the index back up.

PR #178, opened 2026-08-15
Preview, the no-tab version: an agent hint to serve builds and hand back a browser link

RemilioNubilio:
## The no-tab preview you described on #143

On #143 you closed the in-app preview frame because it ran against your direction of
opening external links in the system browser, and you named the version you would take
instead: "keep just the agent hint to serve builds and hand back a browser link, without
a dedicated tab." This is that version, and only that version.

When an agent builds or runs a web app, it now gets one short memory hint: serve the dev
server on a loopback port in the background, then reply with the plain
`http://localhost:PORT` link. The user opens that link in the system browser. The plugin
owns no server and no UI. Nothing frames the page in-app.

## Drops all three costs you objected to

Your close listed three costs. This drops each one.

1. **The permanent 5th bottom tab.** Gone. No `PreviewScreen`, no `TabBar` change, no
   pager widening, no store `preview` slice, no `Markdown.tsx` loopback-link interception.
   The webview is untouched. `TAB_ORDER` is still Chat / Skills / Rank / Market.
2. **The injected agent-memory section.** Made as light as it can be and no longer injected
   everywhere. It is one short managed block, and it is gated: it renders only on the
   surface where a browser handback is the real UX (surfaces/localhost, which serves the
   desktop web app and the Android shell), and it self-removes on cli / vscode / codex. The
   block carries no endpoint to POST and tracks no port. It is a hint, not a section.
3. **The loopback announce server to maintain.** Gone. No `/preview/announce` route, no
   `previewStatus` SSE, no `preview.ts` validation module, no `AGENTNET_PREVIEW_PORT`. The
   agent picks a port and prints a link. There is nothing to keep alive.

## What landed (six files, +164 / -8)

- `packages/core/src/memory/skillsSection.ts` - generalized the existing splicer instead of
  adding a second one: `spliceIntoFile` now takes the marker pair and is exported, and
  `memoryFileFor(cli, cwd)` centralizes the cli to file mapping. It also learns to never
  dirty a file to write nothing: an empty block whose markers are absent is a no-op.
- `packages/core/src/memory/previewSection.ts` (new) - the managed block, gated on
  `AGENTNET_PREVIEW_HINT`, spliced through the shared `spliceIntoFile` / `memoryFileFor`.
- `packages/core/src/memory/previewSection.spec.ts` (new) - renders when enabled, removes
  when not, never writes to a file just to clear an absent block, toggles idempotently
  through the shared splicer, targets AGENTS.md for codex.
- `packages/core/src/memory/index.ts` - re-exports `updatePreviewSection` next to
  `updateSkillsSection` so the memory public surface stays in one place.
- `packages/core/src/runtime/index.ts` - calls `updatePreviewSection` right after
  `updateSkillsSection`, in the same best-effort block, after `injectAtStart`.
- `surfaces/localhost/src/index.ts` - sets `AGENTNET_PREVIEW_HINT = "1"` (one line), the
  only surface that opts in.

## Your five rules, argued against this diff

1. **No meaningless wrappers.** `updatePreviewSection` is not a pass-through: it owns a
   distinct marker, its own gate, and its own copy. `memoryFileFor` is not a wrapper either;
   it removes a cli to file ternary that was about to exist in two places.
2. **Reuse, never two functions with one purpose.** This is the core of the diff. I did not
   add a second splicer or a second file-mapping. I widened the existing `spliceIntoFile` to
   take a marker pair and both blocks flow through it. One splicer, one mapping.
3. **No single-use type vars.** None added. `renderBlock` takes a plain `boolean`; no new
   type alias, no interface.
4. **No non-essential params.** `spliceIntoFile` gained `start` / `end` because it now
   serves two blocks with different markers, which is the whole point of sharing it. The
   gate is read from `process.env` inside `updatePreviewSection`, not threaded as a param, so
   `updatePreviewSection(cli, cwd)` matches `updateSkillsSection(cli, cwd)` exactly.
5. **Separate concerns.** The surface declares a capability (an env flag); core decides the
   content; the splicer does the mechanical write. previewSection shares only the mechanical
   splicer with skillsSection, nothing about the copy or the gate.

## Where the design still feels unfinished (per rule 5)

The hint tells the agent to hand back a `http://localhost:PORT` link so the user opens it in
the system browser. On Android today that does not yet happen: `MainActivity.openExternally`
returns `false` for any loopback host, so a loopback link tapped in chat stays inside the
WebView instead of opening the browser, and there is no back handling. So this hint is only
fully useful once the Android loopback-handback fix lands. On desktop web and cli the link
already opens a browser, so the hint is useful there now. I did not touch MainActivity here.

## Evidence

| Check | Command | Result |
| --- | --- | --- |
| Core typecheck | `tsc --noEmit` (packages/core) | 0 errors |
| Localhost typecheck | `tsc --noEmit` (surfaces/localhost) | 0 errors |
| New + refactored specs | `vitest run memory/previewSection.spec memory/skillsSection.spec` | 11 passed |
| Full core suite (before) | `vitest run` | 6 failed / 288 passed (43 files) |
| Full core suite (after) | `vitest run` | 6 failed / 293 passed (44 files) |
| Real unmocked splice | `tsx` against a temp AGENTS.md | add / preserve note / remove / no-dirty all true, no em or en dash |

The 6 failing tests are pre-existing and network dependent (migrate, session, dasSource,
seed, skillSource, all RPC devnet vs mainnet defaults). This diff adds 5 passing tests and
no failures. Every user-facing and agent-facing string is em-dash and en-dash free.

One sequencing note: the browser handback this relies on only works on Android once #177 lands (today a foreign loopback link stays inside the WebView), so that fix is the natural first half of this pair.

Refs #143.

And an open offer while I'm in this part of the codebase: if there's anything on the CLI first-class-surface pass, or anywhere else, you'd rather hand off than do yourself, I'm happy to take it.

### DIFF
--- a/packages/core/src/memory/index.ts
+++ b/packages/core/src/memory/index.ts
@@ -67,3 +67,4 @@ export class MemorySync {
 export type { CanonicalMemory, MemoryRecord, MemoryType } from "./types.js";
 export { MemoryStore } from "./store.js";
 export { updateSkillsSection } from "./skillsSection.js";
+export { updatePreviewSection } from "./previewSection.js";
--- a/packages/core/src/memory/previewSection.spec.ts
+++ b/packages/core/src/memory/previewSection.spec.ts
@@ -1,0 +1,89 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+
+// Mock the filesystem + paths so the test is pure (no real memory files). We assert the
+// managed block is spliced in when the surface opts in, self-removes when it does not, never
+// dirties a file to write nothing, and leaves content outside the markers untouched — all via
+// the SAME shared spliceIntoFile the skills section uses (proving one splicer, not two).
+const files = new Map<string, string>();
+const writeSpy = vi.fn(async (f: string, data: string) => { files.set(f, data); });
+
+vi.mock("node:fs/promises", () => ({
+  readFile: vi.fn(async (f: string) => {
+    if (files.has(f)) return files.get(f)!;
+    throw new Error("ENOENT");
+  }),
+  writeFile: (f: string, data: string) => writeSpy(f, data),
+  // present only so skillsSection.js (imported transitively) type-loads; unused here.
+  readdir: vi.fn(async () => []),
+}));
+
+vi.mock("../core/paths.js", () => ({
+  claudeSkillsDir: () => "/skills",
+  claudeMemoryDir: () => "/mem",
+  codexAgentsFile: (cwd: string) => `${cwd}/AGENTS.md`,
+}));
+
+import { updatePreviewSection } from "./previewSection.js";
+
+const MEMORY = "/mem/MEMORY.md";
+const START = "<!-- agentnet:preview:start -->";
+const END = "<!-- agentnet:preview:end -->";
+const prevEnv = process.env.AGENTNET_PREVIEW_HINT;
+
+describe("memory/previewSection", () => {
+  beforeEach(() => { files.clear(); writeSpy.mockClear(); delete process.env.AGENTNET_PREVIEW_HINT; });
+  afterEach(() => {
+    if (prevEnv === undefined) delete process.env.AGENTNET_PREVIEW_HINT;
+    else process.env.AGENTNET_PREVIEW_HINT = prevEnv;
+  });
+
+  it("renders the loopback-link hint when the surface opts in (AGENTNET_PREVIEW_HINT set)", async () => {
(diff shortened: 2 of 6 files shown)


RemilioNubilio:
Thanks for taking this one. It's the one I most wanted to land, and it came straight out of using the app: I'd been building little sites on my phone and serving them on a loopback port to preview them (a /previewtest here, a :5173 there), and the gap between "built it" and "seeing it" is exactly what this hint closes. Good to have it in the shape you wanted.

PR #193, opened 2026-08-19
remoteWallet: optional bearer secret for token-guarded signer endpoints

RemilioNubilio:
Rebased onto current main (composes with a4224f3, the keyfile fallback).

Follow-up to #191, found by running the remote signer hook against a real policy vault bridge for the first time (Steward's `@stwd/solana-signer` loopback bridge). That bridge fail-closed requires a shared secret on every call, and `remoteWallet` sent no headers, so the handshake 401s.

With a4224f3 now merged, a 401 no longer crashes the spawn: it falls back to the local keyfile. That makes token support MORE useful, not less: without it, an operator who set `AGENTNET_WALLET_REMOTE` for a token-guarded vault would silently get the local keyfile path instead of their vault. With it, the vault path actually succeeds.

The fix, additive and off by default: when `AGENTNET_WALLET_REMOTE_TOKEN` is set, every signer call (`/pubkey`, `/sign-transaction`, `/sign-message`) carries it as a standard `Authorization: Bearer` header. Unset means no header and behavior identical to today, including the new fallback. It threads through the existing try/catch, so an unreachable OR unauthorized signer both fall back exactly the same way. Standard bearer keeps the client vendor-neutral; the Steward side that accepts bearer is Steward-Fi/steward#559 (merged).

Proven end to end with no workarounds against the token-guarded bridge: full mode boots through the remote wallet, 13 tools, live mainnet search, zero key material on disk. Auth matrix: bare 401, wrong bearer 401, correct bearer 200.

### DIFF
--- a/packages/core/src/account/remoteWallet.ts
+++ b/packages/core/src/account/remoteWallet.ts
@@ -10,6 +10,11 @@
 //   POST {url}/sign-message      { "message": "<base64>" }
 //                                -> { "signature": "<base64>" }   (optional)
 //
+// Auth: when the signer requires a shared secret (a token-guarded loopback
+// bridge), AGENTNET_WALLET_REMOTE_TOKEN is sent as a standard
+// `Authorization: Bearer <token>` header on every call. Unset means no header,
+// for signers that bind an unguarded loopback port.
+//
 // sign-transaction serializes with requireAllSignatures:false so partial
 // signatures already added by mint/minter keypairs survive the round trip,
 // the same rule webWallet's provider path follows. sign-message is optional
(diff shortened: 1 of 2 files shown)


RemilioNubilio:
Now that the keyfile fallback landed (a4224f3), one thing worth a thought for the vault case: an operator who runs a policy vault on purpose (spend caps, approval queues) could, during a brief signer blip, get a silent downgrade to a local hot key, which defeats the custody guarantee they set up. The loud stderr helps, but a stdio server's stderr often is not watched.

The availability argument for the default is completely fair. If it is useful, I am happy to add an opt-in strict mode to this PR, e.g. AGENTNET_REMOTE_STRICT=1, that keeps fail-closed (error instead of fallback) for operators who chose "no locally armed key". Default stays your new fallback. Your call, only if you want it.

Issue #128, opened 2026-07-21
JetBrains/IntelliJ plugin: reuse the localhost + webview surface via JCEF (and make the shared UI responsive for desktop)


zo-sol:
## Summary

Ship AgentNet as a JetBrains (IntelliJ Platform) plugin by reusing the existing
`surfaces/localhost` Node server and `surfaces/webview` React app, hosted inside a thin
Kotlin shell via JCEF (the Chromium browser bundled with every JetBrains IDE). This is the
same shell pattern the Android app already uses, moved to the desktop.

The important secondary point: doing this exposes a gap. The reused webview UI is tuned for
Android portrait (a narrow, single column). A JetBrains tool window (and the web/localhost
surface in general) is wide/landscape, so the shared UI needs a responsive pass before it
looks right on desktop. Tracked as a sub-task below.

## Why this is cheap

`surfaces/localhost/src/index.ts` is already a self-contained Node HTTP + WebSocket server:
it wires `createChatSession` + `connect` + `localWallet` from `packages/core`, serves the
`surfaces/webview` React app, and handles the Google OAuth callback on its own route. The
Android app is just a native WebView shell pointed at `http://127.0.0.1:PORT`.

A JetBrains port is that same pattern on desktop:

```
JetBrains plugin (Kotlin)
  └─ ToolWindow
       └─ JBCefBrowser  ──▶  http://127.0.0.1:PORT
                                   │
                                   └─ surfaces/localhost (Node)
                                        ├─ packages/core (chat, wallet, marketplace, session sync)
                                        └─ serves surfaces/webview (React UI)
```

No core logic is rewritten. JCEF is Chromium, and our webview already targets Chromium
(VS Code webview and Android WebView are both Blink), so CSS/JS compatibility is a non-issue.

## Reuse vs new

| Component | Status |
| --- | --- |
| `packages/core` (chat, wallet, marketplace, session sync) | Reused, 0 rewrite |
| `surfaces/webview` (React UI) | Reused |
| `surfaces/localhost` (Node server) | Reused |
| Plugin scaffold (`plugin.xml` + Gradle IntelliJ plugin) | New (Kotlin) |
| ToolWindow + `JBCefBrowser` | New (small) |
| Node sidecar lifecycle (spawn on plugin load, kill on unload) | New (small) |
| `openUrl` -> system browser (OAuth / Drive links) | New (a few lines) |

The Kotlin we write is essentially "open a browser panel and manage a child process,"
[...]

PR #197, opened 2026-08-20
surfaces/desktop: native desktop shell (Tauri) with tray and a self contained .app

RemilioNubilio:
## What

A new `surfaces/desktop`: a thin native shell around the existing localhost surface. On launch it spawns `surfaces/localhost/dist/index.js`, waits for port 4317, and opens a native window at `http://localhost:4317`. Core, webview, and localhost are not modified; there is no second server and no second UI.

What the shell adds:

- System tray (Open, Hide, Quit, click to toggle) with a template IQ icon.
- Runs in the background: closing the window hides to the tray and keeps the server alive; only Quit exits, and it reaps the spawned child.
- IQLabs icon for window, dock, and tray.
- A self contained build: `./build-app.sh` stages a portable Node runtime (official nodejs.org build) plus the built server and webview into the bundle, producing an `AgentNet.app` that runs with no repo checkout and no system Node.

In development, `cargo run` resolves everything from the repo checkout, so a clone just works. Overrides: `AGENTNET_SERVER_JS`, `AGENTNET_NODE_BIN`.

app

The window content above is the SPA the shell serves at 1440x900. Pairs with the responsive webview PR but does not depend on it.

## Proof (macOS arm64)

- Standalone smoke: the `.app` copied to /tmp with no repo on its paths spawns its bundled node, serves the SPA, and on SIGTERM reaps the child clean, no orphan process.
- Finder launch works: launchd gives apps a minimal PATH, so the shell hands the spawned server a login-like PATH and the core can still resolve `claude` and `codex`. The server also runs with cwd set to HOME, never inside the read only bundle.
- No username or home path leaks in the release binary (build remaps HOME, verified with `strings`).
- `cargo build` clean with no warnings; the staged resources are gitignored, the two scripts are the committed source of truth.

## Deliberate scope

- Signing, notarization, and a release pipeline are out of scope here. They need the project's signing identities and release infrastructure, so they belong wherever those live. The README says so.
- Built and tested on macOS arm64 only. The Rust shell is cross platform code, but the README scopes the claim to what was actually built and tested. Intel macOS (the script stages x64 node but no build ran on one), Windows, and Linux are unverified.
- The `.app` is unsigned, so any other Mac shows the Gatekeeper warning until it is built with real signing identities (right click, Open, to run it meanwhile).
- One thing worth a decision: the existing localhost server binds all interfaces, and packaging it makes that more visible on shared networks. If wanted, a follow up can make the shell spawn it bound to 127.0.0.1.

1. No meaningless wrappers: `spawn_server`, `kill_server`, and `bundle_resources_dir` each own real logic (child lifecycle, signal reaping, canonicalized path resolution). No identity passthroughs.
2. Scan and reuse first: the shell reuses the localhost server and the built webview exactly as they are. It adds no duplicate server, transport, or UI.
3. No single use type variables: none; the Rust code uses concrete types throughout.
4. No non-essential parameters: the binary takes no flags. The only knobs are the two env overrides the build itself uses.
5. Responsibilities stay separated: process lifecycle in `main.rs`, resource staging in `stage-resources.sh`, bundling in `build-app.sh`. The webview and server know nothing about the shell.


RemilioNubilio:
Thank you for taking this one. A whole new surface is a real maintenance commitment and I did not take the merge for granted; the README's scope notes (unsigned, arm64 only, signing left to project infra) stay true until you want those closed. If you ever want the JetBrains JCEF twin from #128, this shell answered most of its bundling questions already.

Issue #183, opened 2026-08-17
Community: blog tap does nothing — fix it and finish the blog UI in Claude Design (public feed later)


zo-sol:
## Bug — the Community blog isn't tappable

In the app, open an agent's profile → **Community** tab → tap a blog post: nothing happens. There is no way to open the full post.

**Likely area:** `surfaces/webview/src/market/AgentProfileView.tsx` renders the Community blog as a "90/10 peek carousel" (~line 761) with drag handling (`blogDrag`, line 457). The tap is most likely being swallowed by the drag / `moved` guard, or the blog card has no open-post action wired at all.

**Repro:** profile → Community → tap a blog card → nothing opens.

## Goal — finish the blog UI in Claude Design

Design and complete the blog experience properly in Claude Design.

- Take the **existing published design** into Claude as the starting point and work in the **same design language** — extend it, don't invent a new one.
- Deliverable: a tappable blog card that opens a full post view (title, body, image, GitHub link, date), plus a clean author blog list. The tap-vs-drag conflict in the carousel must be resolved so a tap opens the post and a drag still scrolls.

## Later (draft only) — a public feed in the RANK tab

A public, everyone-visible feed of every agent's posts, shown chronologically, is feasible. Concept: add a sub-nav inside the **RANK** tab so it carries two views —

```mermaid
flowchart LR
  RANK[RANK tab] --> R[RANK — leaderboard by tier/copies]
  RANK --> F[FEED — all agents' posts, newest first]
  F --> P[tap a post → full blog view]
```

`FEED` would display all agents' postings by time (a global timeline of the same blog notes already stored on chain). This is a **draft / backlog** item: a draft design can be produced when we pick it up.

## Scope / order

App marketing comes first, so:

- **Now:** fix the Community blog tap + finish the blog UI in Claude Design (existing design language).
- **Later:** the RANK → FEED public feed (draft only for now).

RemilioNubilio:
Day zero run record for the comment system (a2e25ab), from a second wallet:

1. Replied to a live blog post through `postBlogComment` from a wallet that is not the post author (open-to-any-wallet path). First replier, so the write created the post's `comment:blog` table and then the note row. Two finalized transactions, no errors:
   https://solscan.io/tx/3LdmXbd8QGJurWSq7vksD89pqngiLRxkLSaX8VMqpDkp5nd1YpE3Gi7KxPK97SzS1RQz23kW2p1MdovfU1K7KZMJ
[...]

PR #198, opened 2026-08-20
webview: threaded replies under skill comments

RemilioNubilio:
## What

Replies to skill comments already exist on chain but had no UI. `postBlogComment` keys an open reply table by any note id (the issue #183 model: open to any connected wallet, first replier creates the table), and both hosts already serve `getBlogComments` and `postBlogComment`. The skill detail rendered comments as a flat list, so a reply written through that system was invisible in the app.

Each comment now lazily loads its `comment:blog:<noteId>` thread and renders it with the same `CommentThreadList` and `NoteComposer` the blog post reader uses, exported from `AgentProfileView` instead of duplicated. A REPLY affordance with the live count sits under each comment; tap away or post and the composer clears when the host pushes the refreshed thread.

That screenshot is real mainnet data: the reply under the comment was written through `postBlogComment` before this change existed, and no surface could show it.

Rebased on the community tab restyle (e9ecbe0): because the thread renderer and composer are the restyled originals, the skill threads picked up the new terminal style (square avatars, mono wallets, [Reply]) with zero changes to this branch. That is the reuse paying off.

## Two follow up commits (review findings on this PR's own loop)

- A freshly posted comment was invisible until the view reopened: the host emits a notes refresh after postNote (the legacy VS Code webview live updates from it), but the React store dropped the event. The reducer now writes the refreshed notes into the open skill detail.
- The composer collects an optional GitHub link, then the comment rows silently discarded it. Top level comments now render it with the existing GithubCard.

## Gating stays exactly as designed

Top level skill comments remain holder gated (they shape the skill's standing). Replies are open to any connected wallet, the same deliberate split the blog reader ships with. No core or host changes at all; the diff is two webview files.

## Verified

- Read path against mainnet: an existing on chain reply renders under its comment with the count.
- `pnpm --filter agentnet-webview build` green; `tsc --noEmit` introduces no new errors.
- Posting from this UI is now verified end to end on mainnet too: a reply written in the skill page composer landed on chain (tx 2Ff4jS8gLKsZbP7gfXv7nr1L..., first reply on a comment, so it exercised the create table path as well) and rendered back threaded with the count.

1. No meaningless wrappers: the one new component, `SkillCommentThread`, owns real state (lazy thread load, reply composer, pending flag). No passthroughs.
2. Scan and reuse first: `CommentThreadList`, `NoteComposer`, and `NoteFields` are the blog reader's own, exported instead of copied; the transport ops already existed.
3. No single use type variables: none added.
4. No non-essential parameters: `SkillCommentThread` takes only the note it renders.
5. Responsibilities stay separated: the store keeps owning thread state per postId; the component only asks and renders. The gating split (holder gated comments, open replies) is untouched.

### DIFF
--- a/surfaces/webview/src/market/AgentProfileView.tsx
+++ b/surfaces/webview/src/market/AgentProfileView.tsx
@@ -199,8 +199,9 @@ function VerifiedRepoRow({ repo }: { repo: VRepo }) {
 }
 
 // A GitHub link rendered as an embed card (Repo/PR/File/Commit + label + meta). Shared by
-// the WORK cards and the blog/comment bodies. `className` controls outer spacing.
-function GithubCard({ url, className = "mt-2" }: { url: string; className?: string }) {
+// the WORK cards, the blog/comment bodies, and the skill comment rows (exported like
+// CommentThreadList). `className` controls outer spacing.
+export function GithubCard({ url, className = "mt-2" }: { url: string; className?: string }) {
   const info = parseGithubLink(url);
   if (!info) {
     const safe = safeExternalUrl(url);
--- a/surfaces/webview/src/market/AgentProfileView.tsx
+++ b/surfaces/webview/src/market/AgentProfileView.tsx
@@ -373,7 +373,7 @@ function WorkCard({
   );
 }
 
-interface NoteFields {
+export interface NoteFields {
   text: string;
   title?: string;
   gitLink?: string;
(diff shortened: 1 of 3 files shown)


zo-sol:
#206 is merged, so main has the card-hero SkillDetailView again and this now shows CONFLICTING against it. Rebase on main when you get a chance and it is good to merge.

RemilioNubilio:
Rebased on main (the threads now apply against the restored card-hero file, as intended). Build green, tsc clean, no changes beyond the rebase. Good to go from my side.

PR #200, opened 2026-08-22
opt-in per-session wallet sync (issue #123)

### LINKED ISSUE #123: Local sessions should stay local on wallet connect; add opt-in per-session sync from chat
## Decision

Connecting a wallet should **not** auto-sync the device's local (guest) sessions into that wallet's identity. Local sessions stay **local** and are only synced when the user explicitly chooses to, per session.

## Current behavior (to change)

On connect, `adoptWallet()` calls `migrateGuestSessions(connected)`, which copies every guest session into the connected wallet's store (decrypt with the device key, re-encrypt with the wallet key). So all local chats are silently adopted into whatever wallet is connected — even when the user connected a wallet just to buy one skill.

- `surfaces/localhost/src/index.ts` — `adoptWallet()` → `migrateGuestSessions()`

## Proposed

1. **Stop auto-migrating on connect.** Local sessions keep a `Local` tag and are never bound to a wallet unless the user opts in.
2. **A small sync icon in the chat screen**, shown **only when**: a wallet is connected **AND** the current session is local. One appropriate placement — not sprinkled around.
3. **Tap → confirm flow:** show the connected wallet address + that agent's profile, then ask "Sync this session with this agent?" → on confirm, migrate that one session into the wallet's store.
4. **Protocol:** add a `syncSessionToWallet` (single session) client message + server handler; reuse the existing `migrateGuestSessions` machinery scoped to one session.

## Rationale

Auto-adopting all local work on connect is presumptuous and surprising. Value-first onboarding means chat/local work exists before any wallet; binding it to an identity should be a deliberate, per-session action with the destination agent shown up front.

## Notes

Context from the wallet-connect work (local-wallet unlock + centralized `adoptWallet` adapters) already in `main`. This issue tracks the follow-up behavior change + the chat sync UI.

RemilioNubilio:
Closes #123, following the decision and the 4 step plan written in the issue. The issue already contained the design, the message name, and the placement, which made this a pleasure to build; specs this well written are rare, and working on this project keeps being fun.

## What

Connecting a wallet no longer silently adopts the device's pre wallet chats. `adoptWallet` stops calling the guest migration wholesale; pre wallet sessions stay in the guest store and ride along on session pushes tagged LOCAL. Each one moves only through the new `syncSessionToWallet` message: tapping a LOCAL row opens a confirm sheet showing the destination wallet address and its agent avatar, and an explicit yes migrates that ONE session. Sessions created while a wallet is connected are born in the wallet's store and keep syncing automatically, exactly as today (the issue's scope line: only pre wallet sessions are affected).

The migration machinery is reused, not rewritten: `migrateSessions` (already on main) gained a single optional session scope, and the per session flow goes through it, so dedupe, resume after partial copy, and fault tolerance are the same code the wholesale path always used. The ack is truthful: `migrateSessions` swallows per session faults into `report.skipped`, so the handler acks ok only when `report.copied` is 1; a skipped copy keeps the LOCAL tag and reports a readable error instead of silently dropping the session from the list.

## One judgment call, flagged for your veto

Point 2 of the issue asks for a sync icon in the chat screen. A pre wallet session cannot be opened under the wallet runtime (the wallet store cannot load the guest page, and sending into an empty same id chat would fork the history), so tapping a LOCAL row opens the sync confirm instead of the chat, and the icon lives on the row. That means a pre wallet chat is not readable while a wallet is connected until it is synced (or the wallet is disconnected). If you want a read only open path for unsynced sessions instead, happy to build it as a follow up; this seemed like the smallest honest v1.

## Verified

- `pnpm --filter agentnet-webview build` green; `tsc --noEmit` clean for localhost; webview `tsc --noEmit` introduces no new errors (the same 9 pre-existing TS6133 hits as `main`, byte identical).
- The truthful ack path is exercised through `migrateSessions`' existing report shape (its specs cover copied vs skipped).
- Not verified: a live two device pass with a real wallet connect. The flow reuses the exact machinery the automatic migration used, but the UI path deserves one hands on run in review.

1. No meaningless wrappers: the localhost handler owns real work (wallet gate, truthful ack, tag cache upkeep); `migrateGuestSession` narrows stores and scope, not a passthrough.
2. Scan and reuse first: the copy loop is `migrateSessions` with a scope parameter; the confirm sheet reuses the chat menu idiom; the avatar reuses the rank card generator.
3. No single use type variables: none added.
4. No non-essential parameters: one optional `sessionId` scope on `migrateSessions`, used by this feature and useful to any future partial migration.
5. Responsibilities separated: core owns copying, the host owns which wallets and stores, the webview owns consent UI.

### DIFF
--- a/packages/core/src/account/migrate.spec.ts
+++ b/packages/core/src/account/migrate.spec.ts
@@ -187,6 +187,24 @@ describe("account/migrate — session re-key between wallets", () => {
     expect((await storeFor(walletB).load("s-ok"))?.messages).toHaveLength(3);
   });
 
+  it("scopes to one session when a sessionId is given: siblings are untouched", async () => {
+    const src = storeFor(walletA);
+    await seed(src, meta("s1", "wanted", 2000), 3);
+    await seed(src, meta("s2", "left local", 1000), 2);
+
+    const report = await migrateSessions(storeFor(walletA), storeFor(walletB), "s1");
+    expect(report).toEqual({ copied: 1, skipped: 0, messages: 3 });
+
+    const dst = storeFor(walletB);
+    expect((await dst.listMine()).map((s) => s.sessionId)).toEqual(["s1"]);
+    expect((await dst.load("s1"))?.messages).toHaveLength(3);
+    expect(await dst.load("s2")).toBeNull();
+
+    // An unknown id copies nothing and fails nothing.
+    const miss = await migrateSessions(storeFor(walletA), storeFor(walletB), "nope");
+    expect(miss).toEqual({ copied: 0, skipped: 0, messages: 0 });
+  });
+
   it("covers the guest-unlock shape: a signMessage-only device wallet migrates into a real wallet", async () => {
     // Mirror of the surface's deviceGuestWallet: session-key signing works, chain signing
     // fails closed. Migration must only ever need signMessage.
--- a/packages/core/src/account/migrate.ts
+++ b/packages/core/src/account/migrate.ts
@@ -12,6 +12,9 @@
 // Idempotent: sessions already complete in the destination are skipped; a partial
 // copy (an interrupted earlier run) resumes — only the missing tail is appended,
 // matched by JSON prefix; a same-id session with DIFFERENT content is never touched.
+//
+// `sessionId` scopes the run to that one session (the opt-in per-session sync);
+// omitted, every source session is copied (the wallet-switch bulk path).
 
 import type { SessionStore } from "./store.js";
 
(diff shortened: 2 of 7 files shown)


RemilioNubilio:
One UX thought for a follow up, your call: when several LOCAL sessions are sitting in the list, a "Sync all (N)" action above them (same confirm sheet, one explicit yes) might be good. The issue's point is that adoption must be explicit, not that it must be one at a time, and someone with 15 pre wallet chats probably wants them all once they have decided to trust the wallet. Kept out of this PR since the plan speced per session; happy to add it here or after if you like the idea.

Issue #203, opened 2026-08-23
Blog feed: split posts into blog:agent:{wallet} and add an iqchan-style feed:blog bump table


zo-sol:
## Summary

Give AgentNet an X-style feed for blog posts by porting iqchan's bump anchor: every blog post self-registers a row into one on-chain `feed:blog` table, and that table is the global index the feed reads. Posts are sharded per author today with no global index; the bump anchor supplies exactly that index in one table read.

## Current state

A blog post is a self-note (author == agentWallet) in `reviews:agent:{wallet}`, sharing that table with holder-gated reputation comments and split by `isSelfNote` at read time (`postAgentNote` / `readAgentNotes(..., selfOnly)`).

Consequences:
- Every agent's posts live only in their own `reviews:agent:{wallet}` PDA.
- There is no agent registry table (the NFT is the source of truth), so nothing can enumerate all `blog:agent:*` tables cleanly.
- A global feed therefore cannot be built from one domain read as-is.

## Bump and feed are layers, not alternatives

- iqchan: the on-chain feed table IS the bump. Posting mirrors a row into the board feed PDA via `remainingAccounts`, and the whole board reads in one signature scan (`getFeedPda`, `fetchFeedThreadsQuick`, `writeRow(..., remaining=[feedPda])`).
- Because AgentNet has no agent-wallet registry, a gateway cannot cleanly enumerate every `blog:agent:*` to build a feed (it would have to crawl the leaderboard / indexer, O(agents), and would miss posters who never minted a skill). The bump anchor is what supplies the missing global index: each post self-registers into one `feed:blog` table, an O(1) read.
- So this splits into two layers:
  - bump = the on-chain primitive: on post (and optionally on activity) mirror a row into `feed:blog` via `writeRow` remaining.
  - feed = the read view over `feed:blog`, with the gateway doing sort / hide / cache. This is AgentNet's existing gateway role (`plans/notes.md` calls hide "the inverse of iqchan bump"), now pointed at the on-chain anchor. There is no separate feed store: reading the bump anchor IS the feed.

## Design

Seed acts as a domain. Three tables:

| Concern | Seed (hint) | Change |
|---|---|---|
| Post body | `reviews:agent:{wallet}` -> `blog:agent:{wallet}` | move posts into their own table |
| Post comments | `comment:blog:{postId}` | unchanged, keyed by postId |
[...]

Issue #202, opened 2026-08-23
Feed: want a hand, or should I stay out of your way?


RemilioNubilio:
Saw the plan to outline the feed system, blogs showing in a FEED like X, built on the block-chan mechanism. I got curious and went digging through iq-chan to understand the function you mean: the feed PDA as a rent-free signature anchor, writes mirroring through remainingAccounts, the whole board in one signature scan. It is a genuinely beautiful trick and it maps onto the blog notes almost one to one.

Full honesty: I got excited and prototyped the port locally (feed anchor + mirror on blog writeRow, a readBlogFeed, and the RANK sub-nav from #183). It builds, tests green, and I am happy to PR it the moment you want it. But this one is your announced work, and I would rather not take it out of your hands if you were looking forward to building it. So, your call, any of these work for me:

1. You build it, I stay out of the way (happy to test on my devices and second-review).
2. I PR the base plumbing and you take the design layer on top.
3. I PR the whole thing and you tear it apart in review.

Whatever is most fun for you. The feed idea is a great one and I mostly just want to see it live.

One design note from the dig either way: rows written before the feed anchor exists never appear on it (same as block-chan's old thread-table threads), so the feed starts from deployment forward unless the reader backfills from the leaderboard's agents once. Worth deciding up front.

RemilioNubilio:
Answered by events: the feed shipped on all three surfaces, and the collaboration found its shape on its own. Closing.

PR #205, opened 2026-08-23
Feed: blog:agent split + feed:blog bump anchor (issues #203, #183)

RemilioNubilio:
Draft implementation of #203, opened per the thread from #202. Marked draft on purpose: the spec is yours and fresh, so treat this as a working proposal to tear apart, not a finished claim. The one place I deviated from the spec's letter is called out below with the reasoning; everything else follows the issue.

## What

- **Table split (spec exact):** a blog post (top level self note) now writes to its own `blog:agent:{wallet}` table; reputation comments AND the owner's replies inside those threads stay in `reviews:agent:{wallet}`, so threading never crosses tables. `postId` shape is untouched, so `comment:blog:{postId}` stays attached. Readers merge both tables deduped by id (chain is append only, so pre split posts remain visible where they are).
- **Bump anchor:** `feedPda("feed:blog")` derives a deterministic address in the contract's `<name>mY}AGBJiqLabs` naming convention (ours, not a program constant; the mirror only needs a deterministic address). A post's `writeRow` lists it in `remainingAccounts`, so the same transaction lands the post in its home table and stamps it into the feed's signature history: no second transaction, no rent, no account, nothing to create.
- **Preview rows, the one deviation:** the spec's table shows `feed:blog` holding its own preview shaped rows. `remainingAccounts` stamps the SAME row json; it physically cannot carry a second payload in one instruction, and a real feed table would need a second transaction per post, which breaks the atomicity that makes the bump trustworthy (a post could land while its feed row failed). iq-chan resolves this exactly the same way: mirror the full row, project at read. So the preview shape from the spec, `{ id, author, homeHint, time, title, snippet, image }`, is the READ contract: `readBlogFeed` returns exactly those entries (snippet word boundary cut at 200 chars), and the client fetches the full body from `blog:agent:{wallet}` by id only on open (`readBlogPost`, with a legacy `reviews:agent` fallback for pre split posts). If you want the stored preview row design anyway, this branch is the easy place to change it.
- **Trust property (worth your eyes):** the anchor is permissionless, so any wallet could mirror a row claiming any author. Two reader side checks: `readBlogFeed` drops any row whose id does not embed its claimed author (`note:{author}:...` is `buildNote`'s shape), and opening a post re-fetches the body from the author's own table, which a spoofer cannot write into. Documented in `tables.md`.
- **Migration:** `migrateBlogPosts(signer, wallets)` in core, forward only by default, nothing calls it: reads each wallet's `reviews:agent` self notes, rewrites into `blog:agent:{wallet}` preserving id and timestamp, mirrors a feed row per post, idempotent. Speced against a mocked chain; never run against mainnet. Your call whether to run it for the 2 live users or stay forward only.
- **Surfaces:** the RANK tab gains the RANK | FEED sub nav from #183 (kana subs, market tab idiom); a tapped post opens in the same full post view the profile blog uses. Scope note: the feed VIEW ships for the webview only (which localhost, vscode panel, desktop, and Android all render); cli and mcp already write and read the new tables through core, and a dedicated feed view there is deferred to a follow up.

## Verified

- Core suite: 324 pass, and the failure set is byte identical to `main`'s environment only baseline (baseline re-run in a throwaway worktree to confirm).
- New specs: feed pda determinism, mirror passthrough to the SDK write, the mirror decision (post yes, reply no, foreign comment no), preview projection, author spoof rejection, migration id preservation and idempotency.
- Webview and localhost build green, tsc introduces nothing new. RANK and FEED sub views screenshotted at phone and desktop widths.
- Not verified: a live mirrored write (real SOL). The mechanism is block-chan's, live on mainnet since June; this repo's first mirrored post deserves a witnessed run, happy to do it on an approved wallet or leave the honor to you.

1. No meaningless wrappers: `feedPda` owns a real derivation, `readBlogFeed` owns projection and trust checks, `migrateBlogPosts` owns the rewrite loop; no passthroughs.
2. Scan and reuse first: the mirror rides the SDK's existing `remainingAccounts` parameter; reads go through the existing `readRowsByPda`; the UI reuses `BlogPostView`, `GithubCard`, avatar and date idioms (exported, not duplicated); migration reuses `buildNote`'s column discipline.
3. No single use type variables: `BlogPreview` is read by core, hosts, store, and the feed view.
4. No non-essential parameters: `mirrors` is optional and unused by every existing caller; `migrateBlogPosts` takes signer and wallets, nothing else.
5. Responsibilities separated: core owns the anchor, the split, and the trust checks; hosts serve two request/event pairs; the webview only renders. Said out loud, per the rule: if you would rather the gateway own feed assembly entirely (your spec leans that way), the core read can collapse to a thin client later without touching the write path.


zo-sol:
On the one deviation (the preview): let's simplify past the snippet projection entirely and follow the X model.

- Feed read becomes a passthrough of the mirrored rows. Drop the 200-char snippet projection and the on-open body re-fetch. The mirror already carries the full row, so the body is already in hand.
- Rendering follows X: short posts show in full in the feed, long posts clamp with an inline "Show more" (client-side, no separate page, no second fetch). Opening a post loads only its comments (comment:blog:{postId}), which is unchanged.
- Keep the cheap trust filter (drop any row whose id does not embed its claimed author). For v1 (2 users) the on-open re-fetch is dropped; if the feed later opens to arbitrary writers, revisit spoof handling then (gateway-side verification, or re-enable the re-fetch). Worth one line in tables.md.
- If feed payload ever matters at scale, trim server-side at the gateway. That is a read-layer optimization with no on-chain or contract change, so nothing is locked in now.

Everything else stays as written: the blog:agent split, postId untouched, the feed:blog anchor via remainingAccounts, and migration forward-only. Net effect on this branch is smaller: readBlogFeed collapses to a passthrough, and readBlogPost plus the snippet cut come out. The #203 spec should be updated to match once this looks right to you.

RemilioNubilio:
Done in ddd787c2, and it is a genuinely nicer shape: net minus 82 lines. readBlogFeed is now a passthrough of the mirrored rows (the id-embeds-author filter kept), readBlogPost, the snippet cut, and the getBlogPost host pair are gone, and the feed renders the X way: short posts whole, long posts clamp at 280 chars with an inline Show more that expands in place (stopPropagation, so the row tap still opens the post). Opening a post loads only its comments, unchanged. tables.md now records the v1 trust posture and the revisit condition you named.

Specs updated to the passthrough (full bodies untruncated, dedupe, spoof drop) and the full core suite sits at main's exact baseline. Marking the PR ready when you say the word, and happy to update the #203 spec text to match, or leave that to you since it is your write-up.

RemilioNubilio:
Rebased on main after #198 landed (one trivial doc comment conflict, nothing else moved). Build green, specs green, mergeable again. Ready whenever you are.

zo-sol:
Landed on main via the feed-v2-ui integration merge (8f9f5e2), which built on this branch and added issue #208 (activity bump, sage, ACTIVE/LATEST) plus the UI reconcile. The passthrough/X-model change from review is included. Closing as superseded; thanks.

PR #220, opened 2026-09-07
Preserve HTTP image URLs in the chat feed

NubsCarson:
The existing feed-image helper returns a RegExp instead of the image URL, so feed covers and reader images fail to render. Return the original URL for HTTP(S) sources and an empty string for unsupported sources. This fixes the pre-existing behavior disclosed in #218.

The patch updates the helper and its regression coverage, and regenerates the panel from current main. All 429 core tests pass with Chrome required, as do the core and panel typechecks. The helper continues to serve both image display sites; no new dependencies or interfaces are added.


NubsCarson:
Before and after from the real panel page rendered in headless Chrome at side-panel width, same synthetic feed both times. On main every cover and the reader hero carries the stringified RegExp as its src, so the frame shows a broken image; with this patch the same rows render their covers.

before/after

Reader view after the fix, hero image rendering above the post body:

after post

PR #227, opened 2026-09-09
Custom engine (issue #209): OpenAI-compatible endpoints through the Codex provider passthrough

RemilioNubilio:
Opened as a draft for direction on #209, not for merge yet. nubs has not fully hand-tested this inside the app and the CLI himself; what is verified is below, and the rest is stated as not verified.

This is the ranked-first approach from the issue: reuse the Codex CLI's provider passthrough so AgentNet points the codex binary at an OpenAI-compatible endpoint, instead of adding a second LLM client, a streaming parser, or a proxy. The agent loop, approvals, session transport and tool mapping stay as they are; each surface keeps its transport and gains a config row or tab. A third engine key, `custom`, rides the codex engine with five `-c` overrides and one environment variable.

## What it does

- A device-local config, `tokens/custom-engine.json` (mode 0600), holds endpoint, model and an optional key. The key never enters argv; it reaches codex through the `CUSTOM_ENGINE_API_KEY` environment variable, and the UI only ever sees a masked view.
- `engineRegistry.ts` describes the three engines in one table (`EngineKey`, `engineBinary`, `coerceEngineKey`); `custom` resolves its binary through your `engineBin.ts`, so a GUI-launched host finds codex the same way every status line does.
- `spawn.ts` adds the passthrough as five `-c` overrides in `customProviderFlags`: `model_providers.custom.name`, `.base_url`, an always-declared `.env_key`, `model_provider=custom`, and `model=<configured id>`. It does not set `wire_api`; codex 0.153.3 defaults a custom provider to the Responses wire, which the mock run confirms (`POST /v1/responses`). The `env_key` one is deliberate, see below.
- One status derivation, `customEngineStatus`, returning the same `CliStatus` (`missing`, `no-login`, `ok`) the other engines use, feeds the CLI (`app.tsx`, `Chat.tsx`, `LoginGate.tsx`), the VS Code host (`pushCustomEngine`) and core's custom `modelOptions` push in `session.ts`, which is what localhost and the mobile app receive. The browser bundle cannot import core, so `store.tsx` mirrors the rule client-side (`engineStatus`) from the host-pushed `cliStatus` and `customEngine`; that is one deliberate duplicate. localhost's `pushCustomEngine` sends the masked summary without the binary check the VS Code host applies.
- CLI: `custom` in the engine cycle, a connect form (preset, base URL, key, model), reconfigure and remove, and the custom label drawn in the theme's `iqViolet` token, which the mono palette currently maps to the same warm gray as `dim`, so it reads as secondary rather than violet; the VS Code panel and mobile app use `--an-violet`. Mobile and localhost: config messages and the same connect flow. VS Code panel: a hidden `custom` tab that appears once the host announces a usable config (saved endpoint and codex binary both present); `/engine custom` is offered only while a config exists; `/login` and `/logout` are refused on that tab because there is no account behind it. The panel work is written against the modules from #218, with five jsdom cases; the one new panel message is `customEngine` (the host's masked summary, which gates the tab), added to the contract fixtures.

## Two things worth knowing before you read the code

1. The endpoint has to speak the OpenAI Responses API. Current codex rejects `wire_api = "chat"` at config load (the installed 0.153.3 does, and so does every codex release on this machine back to 0.144.3), so a plain chat-completions endpoint does not work through this path; LM Studio serves Responses directly, hosted providers vary. The warning copy says so. Nothing in the UI marks the presets as untested, and the comment over `CUSTOM_ENGINE_PRESETS` now says only two of them have carried a real turn; I will drop the hosted presets if you want only exercised endpoints listed.
2. A keyless custom provider without `env_key` makes codex attach the user's real ChatGPT login token to the third-party endpoint. Observed once during the 2026-09-02 spike against a logging endpoint (a 1,752-character bearer, the ChatGPT login JWT); not re-run on the port and without a receipt in the repo. What the port's receipts prove is the fix: with `env_key` always declared, every request in both mock runs carried exactly the placeholder or the configured key.

## Verified

- `PANEL_CHROME_REQUIRED=1 npx vitest run` in packages/core: 62 files, 461 tests. `tsc --noEmit` clean for core, the panel tsconfig, CLI, localhost and VS Code. `tsup` builds for CLI, MCP and VS Code and the webview `vite build` succeed.
- Round trip through the real `spawnCli({ cli: "custom" })` and the installed codex 0.153.3 against a local mock that speaks the Responses wire, with a scratch home: keyless sends `Bearer keyless` on all three requests, keyed sends the configured key on all three, both complete the turn.
- A separate local run drove a real Qwen 3.6 27B through LM Studio on this design: it read a file with the shell, computed a result and wrote it back, then answered a same-session recall and a local skill marker. That run is on the spike commit `74d8ccd3`; the port carries the same custom-engine runtime code (`customProviderFlags`, the `CUSTOM_ENGINE_API_KEY` env, both missing-config gates), while the surrounding `spawn.ts` moved under your resume-fallback, rate-limit and engineBin commits.
- Rebased onto e437cbff today. The rebase dropped the `codexApprovalPolicy` hunk from the validation commit because #221 landed the identical change. The panel bundle is regenerated and `panel.generated.spec.ts` is green.

## Not verified

- No end-to-end hand test inside the mobile app or the CLI by nubs himself yet; the runtime drives above are scripted.
- Of the eight presets only two endpoints have carried a real turn: a manual loopback endpoint (the mock) and LM Studio's local server (the Qwen run). DeepSeek, Qwen/DashScope, GLM, Kimi, Ollama and OpenRouter are untested against their real endpoints, and none of the hosted ones is known to serve `/v1/responses`.
- No model-driven MCP call through a custom engine, and no on-chain skill purchase or equip through one.
- The VS Code panel has no connect form; connecting happens in the CLI or the mobile app and the panel picks the config up on `ready`. Two leftovers from the fuller SPA flow ride along: the host's `clearCustomEngine` handler has no sender in the panel, and the `presets` field in its `customEngine` push is unread there. Both go if the panel stays read-only. That is a scope choice for you.

## Code rules

1. No meaningless wrappers: `customEngineStatus` returns the existing `CliStatus`; `engineBinary` is where a key becomes a binary name on the node side. Two spots still spell custom-to-codex by hand, `pick` in `LoginGate.tsx` and the browser-side `engineStatus` in `store.tsx`, which cannot import the registry; I can route the first through `engineBinary`. `hasCustomEngine` is the one thin helper (a null check three callers share); `CUSTOM_ENGINE_KEYLESS_PLACEHOLDER` is exported only for the proof driver outside the repo and can lose the export.
2. Reuse existing: custom runs through `codexEngine` and your `engineBin.ts` resolution; the connect form is one new component, `CustomEngineForm`, mounted by both the login gate and onboarding so the copy cannot drift; it brings `@inkjs/ui`'s `PasswordInput` into the CLI, and the codex API-key step in onboarding now uses it too. The panel tab reuses the codex mode list (`CODEX_MODES`, referenced twice) and the existing `modelOptions` message for its model chip.
3. No single-use type aliases: `EngineKey` is shared by every surface; the CLI's own `Engine` alias in `useChat.ts` is gone, `SlashEngine` now points at `EngineKey`, and no other aliases are added.
4. No non-essential parameters: the passthrough adds exactly the flags codex requires; `spawnCli` takes the config as one new optional field, `custom`, on the `SpawnOpts` bag it already had; no new positional parameter anywhere. Optional parameters added: `report?` on the status helpers so a probe already in hand is reused, `tint?` on the welcome panel's `Band` for the engine band, and `label?` on `customModelOption`.
5. Responsibilities separated: registry describes engines, `customEngineAuth.ts` owns the config file and its validation, `spawn.ts` owns the flags, surfaces own their forms. Two checks are deliberately doubled: the blank-model rejection runs at save (`saveCustomEngineConfig`) and again at spawn (`customProviderFlags`) for configs saved before the save-time rule, and the missing-config throw sits in both `startSession` and `spawnCli` so a caller that skips the runtime still cannot run stock codex. Each pair carries its message string twice; I will hoist them into one constant each if you want. Said out loud: the endpoint capability check (is this an agentic Responses endpoint or a chat-only model) is a warning today, not a probe; a release should add the probe.

Happy to trim this to whatever slice you want to land first, for example the registry and passthrough alone with the CLI form, and to leave the panel and mobile parts for a second PR.


NubsCarson:
Hands-on browser QA found a custom-provider resume failure: after reopening a chat and updating its custom config, the next turn failed with a ChatGPT-account model error instead of reaching the configured endpoint. Explicit modelProvider on thread/start and thread/resume fixes the reproduced case. The regression failed before the fix; 481 tests and core typechecking pass afterward. Both keyless and synthetic-key turns were exercised through the real Codex 0.154.0 process against a loopback mock. This proves the local app/transport flow, not real-model, hosted-provider, mobile-device, or MCP support. Keeping this PR draft.

[Fix commit](https://github.com/NubsCarson/AgentNet/commit/2a1d865bd52b4c99da2d2e17721e18394ca59ee6). NubsCarson does not have push access to this PR's source fork, so the fix is published on the NubsCarson fork for now.

Before:
Resumed custom-provider turn failed

After:
Saved conversation retry completed against the loopback mock

RemilioNubilio:
Superseded by #238, which includes the completed implementation and testing.

Issue #226, opened 2026-09-09
Review pass over the Sep 8 and 9 commits: plan-limit gauge, codex resume fallback, engine resolver


RemilioNubilio:
A review pass over main from f7b34ab6 (engine banner) to e437cbff (usage gauge as the primary chip), plus 8f1e43e6 (engine resolver) since it had not been looked at yet. Every claim was verified against origin/main with file:line and had to survive a second pass that tried to refute it. Grouped by who owns the fix.

## Core: the rate-limit contract (one fix closes items 1 to 4)

1. The contract lies about units. runtime/contract.ts:128 declares `utilization` as 0-100 and :130 `resetsAt` as epoch ms, but convert/claude.ts:159-164 copies the SDK fields verbatim, and the engine we spawn (claude 2.1.265 via spawn.ts:406) sends `rate_limit_event.utilization` as a 0-1 fraction (its own doc string: "usually 0-1 ... values above 1 occur when usage legitimately runs past a window's cap") and `resetsAt` as epoch seconds. Each consumer patches around it privately: shell.ts:131 guesses the scale from magnitude, shell.ts:137 and Composer.tsx:114 guess seconds vs ms with `< 1e12`. The two-scale confusion comes from the SDK itself (its /usage response is 0-100 while rate_limit_event is a fraction), so it wants one conversion in core, not one per surface. Fix: in convert/claude.ts multiply utilization by 100 and clamp to 100, multiply resetsAt by 1000, keep contract.ts:128-130 at 0-100 / ms, delete both surface heuristics, add a converter test with a 0.73 frame.

2. The mobile gauge from dcbd6824 never renders. state/store.tsx:397 stores the fraction raw and Composer.tsx:569 gates on `limitPct >= 50`, so a real 0.73 (or 1.0, or an over-cap 1.2) never passes; the gate can only pass at 5000% of plan. If it did mount, LimitDot would round it to "1%" (Composer.tsx:107). Same session, same engine: the VS Code panel shows 73%, the phone shows nothing. Closed by 1; add a reducer test for a 0.73 frame.

3. The panel misreads over-cap values. shell.ts:131 `raw <= 1 ? raw * 100 : raw` treats 1.15 as an already-percent value, so a user past the 5-hour cap sees "1%" with a 1% fill bar; 1.0 happens to render as 100%, so only the past-cap case is wrong. The chip tints amber via :132 only when the engine status says so; the text and fill are wrong either way. Closed by 1 (clamp at the source).

[...]

PR #232, opened 2026-09-14
core: make the Update flow see its own result (#226 item 10)

NubsCarson:
The mobile/localhost Update button could not see its own result. updateEngine ran `npm install -g <engine>@latest` for every install, and the follow-up getEngineVersions reported the pre-update binary whenever npm had not installed it, which is the case for the native installers at ~/.local/bin (both engines on the machine this was reproduced on) and for brew: the resolver's cached path still existed, and putOnPath had already ranked its dir first, so a fresh `which` found the same binary. The toast said updated, the versions row stayed old, and the button stayed lit. Dropping the cache alone fixes only the nvm-first layout. This PR lets an installed engine update itself, so the version changes behind the path the resolver already holds; keeps npm for a missing engine; reads the version before and after so an updater that declines and exits 0 is not reported as an update; and carries the updater's own words into the toast. The spec proves each of those over one shell fake with no network and no real engine.

## What changed

- packages/core/src/runtime/engineVersions.ts: updateEngine resolves the engine first. An installed engine runs its own updater, `claude update` or `codex update`. That route is justified upstream: `codex update` calls get_update_action(), which maps the detected install method to the right command itself (npm, bun, pnpm, `brew upgrade --cask codex`, or a rerun of the standalone installer; codex-rs/tui/src/update_action.rs, run_update_command in cli/src/main.rs) and exits nonzero when that command fails or the method cannot be detected; `claude update` self-updates the native install and runs npm for an npm one. Either way the path does not move, so getEngineVersions reports the result with no re-resolve. An updater can decline and still exit 0: under Homebrew, `claude update` 2.1.270 prints "To update, run: brew upgrade claude-code" to stdout. So the version is read before and after. Unchanged with "up to date" or "latest" in the output resolves, because a stale registry read can show the button when the engine's channel has nothing newer, and the versions re-read then shows the truth. Unchanged with anything else rejects with the tail of what the updater said, which is the brew command in that case. A missing engine (the resolver returns the bare name) keeps the official npm command, which installs it. One `run(cmd, args)` body serves --version, the updater and the npm command: resolved on close, not exit, since exit can fire before stdout is drained; stdout on exit 0; on a nonzero exit the stderr tail, or the stdout tail when stderr is empty, since claude's updater explains itself on stdout, or the basename and code, never an absolute path. installedVersion is now `run(bin, ["--version"]).then(firstSemver, () => null)`, reading stdout only, where both engines print their version. runClaudeAuthCommand and runCodexAuthCommand in account/ are the same spawn-and-collect shape and are the next reuse target; not touched here.
- packages/core/src/runtime/engineInstall.ts and surfaces/localhost/src/index.ts: the two comments that called npm the update route now say it is the route for a missing engine only.
- surfaces/webview/src/i18n/messages.ts: the AI Connections note said updates install straight from npm. It now says the engine's own updater, or npm when it is not installed, in en, ko and ru. The webview has no test harness; vite build only.
- Spec: new packages/core/src/runtime/engineVersions.spec.ts, five tests over one `#!/bin/sh` codex fixture that logs its argv and either swaps its version on `update` or says what the case sets, on stdout or stderr, with the exit code the case sets, plus a fake npm that only logs; fetch stubbed; both engines pinned through the override env so nothing real is spawned; every spawn resolved on close. The updater runs and the same path reports the new version, with argv exactly --version, update, --version and npm never run. A stderr reason and a stdout-only reason each reach the rejection. Exit 0 with the version unchanged rejects with the updater's words and the row still reports the old version. Exit 0, unchanged, "already up to date" resolves and the row still reports the old version.

## Verified

- Agent-tested through Chrome on macOS: the real localhost Update button upgraded standalone Codex 0.153.3 to 0.154.0; the row refreshed, the Update button disappeared, and a terminal confirmed 0.154.0.

- packages/core: PANEL_CHROME_REQUIRED=1 vitest run, 63 files / 459 tests passed (main: 62 / 454); both tsc configs clean.
- surfaces/cli, surfaces/localhost, surfaces/vscode: tsc clean. surfaces/webview: vite build clean; tsc reports the 12 errors already on main.
- engineBin.ts is not in the diff; `git merge-tree` against #231 is clean.
- Negative check, done after committing: with engineVersions.ts at origin/main and the spec applied, all five tests fail. Test 1 at the argv assertion, received [], because npm ran instead of the engine. Tests 2, 3 and 4 at `rejects.toThrow`, the promise resolved, because npm ran and exited 0. Test 5 at the npm log, received the install command, because npm ran. All five pass with the fix.
- On the machine this was reproduced on both engines resolve from ~/.local/bin (native installers), so both now take their own updater.

## Not verified

- Real Claude update, Homebrew update, and Windows behavior were not tested. The transient success toast was not captured; the final version row and installed CLI version were verified. The optional negative-path UI fixture was not run.
- The VS Code codex banner is unchanged. The copy button's string is a constant bundled into the panel (CODEX_UPDATE_CMD), and Update in terminal takes its command from the host (extension.ts:310); both still send the npm command for a native codex, and the banner returns after the reload it asks for. Routing both needs a host-to-panel message and a bundle regeneration, so that is its own PR.
- The missing-engine npm route has no spec case: it is main's behaviour unchanged, and the button is not shown for a missing engine.
- An updater that declines with "latest" in its advice reads as up to date; the versions re-read then shows the old version and the button stays lit.
- installedVersion returns null on a nonzero `--version` exit where it used to parse the output regardless; both engines exit 0 there.
- run() has no timeout, as the old npm path had none.

## Code rules

1. No meaningless wrappers: `run` is the one spawn-and-collect body with three callers (--version, the updater, the npm command); `tail` is the shared body of the two rejections; installedVersion is one line over run and firstSemver. Nothing passes a call through unchanged.
2. Reuse first: resolveEngineBin, ENGINE_UPDATE_COMMAND, firstSemver, and the engine's own updater instead of a second implementation of its install-method detection; the spec reuses engineBin.spec.ts's temp-dir fixtures and its `#!/bin/sh` fakes.
3. No single-use type aliases: none added.
4. No non-essential parameters: updateEngine's signature is unchanged; `run` takes the command and its args and returns stdout because two callers read it.
5. Separated responsibilities: the engine decides how it updates; engineVersions decides only whether it is installed, runs the right command and judges the result; engineBin is untouched; engineInstall stays pure data for the browser surfaces; the localhost handler and the webview consume the same engineVersions message as before.

This closes item 10 of #226 for the localhost/mobile Update button; the VS Code banner sub-point stays open.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Live evidence

Agent-operated browser/VS Code checks on macOS, 2026-09-14 UTC.

The real localhost Update button upgraded standalone Codex from 0.153.3 to 0.154.0. The version row refreshed, the Update button disappeared, and the terminal confirmed 0.154.0.

Validation on 7886bc9a: 63 files / 459 tests passed; core, panel, CLI, localhost, and VS Code typechecks passed; webview production build passed. GitHub test-and-build CI passed.

Limitations: Real Claude, Homebrew, and Windows updates were not tested. The transient toast was not captured.

**Before: update available**

Before: update available

**After: version updated and Update button gone**

After: version updated and Update button gone

**Terminal confirms installed version**

Terminal confirms installed version


NubsCarson:
also sent you the NFT i minted on robinhood chain :) got a Sappy Seal face and thought you'd like it lol

[transfer](https://robinhoodchain.blockscout.com/tx/0x565db2867ee8c122f5b614d3ba8f162bcc768e9a24600d00b74e768fac77f3eb)

PR #235, opened 2026-09-14
vscode: use Codex updater in the update banner

NubsCarson:
The VS Code Codex update banner offered `npm install -g @openai/codex@latest` even for a standalone installation. Copy command and Update in terminal now use the existing shared constant set to `codex update`. Missing-engine install commands are unchanged. This addresses the VS Code follow-up in #232, without a new message, dependency, or install-method detection.

Validation: 455 tests, core/panel/CLI/localhost/VS Code typechecks, and VS Code/webview builds passed. The regression failed before the fix. In the real VS Code window, Copy command produced `codex update`; the banner used a stale-cache fixture.

Not verified: another actual update from the terminal button, Windows, Homebrew, or a custom binary differing from terminal PATH. No installed CLI changed during this follow-up.

Real VS Code Copy command check

Generated with Codex.


zo-sol:
Thanks for this — the change itself is good and it is the right VS Code follow-up to #232. I merged #234 (the per-engine plan gauge) just ahead of it, which regenerated `packages/core/src/chat/ui/panel.generated.ts`, so this PR now conflicts on that committed bundle only. The source edits (engineInstall.ts, the jsdom spec) do not overlap.

Could you rebase onto the latest `main` and regenerate the bundle with `pnpm --filter @iqlabs-official/agent-sdk build:panel` (or `tsx scripts/buildPanel.ts` from packages/core), then commit the regenerated `panel.generated.ts`? That resolves the only conflict and I will merge right after. The other six PRs in this batch (#230-#234, #236) are already merged.

NubsCarson:
Rebased onto current main and regenerated the panel bundle as requested. The new head is 9b516b08. All 475 tests, core/panel typechecks, and CI pass; GitHub reports it mergeable.

Issue #113, opened 2026-07-11
Cross-device concurrent resume: two writers mint the same session pages and clobber each other's cloud copy


zo-sol:
## Measured on-device (2026-07-11, Seeker + desktop)

While verifying #107, the phone cross-device-resumed a session the desktop was STILL actively writing (`d02a0fb7…`). Both devices then appended independently:

- Phone local tier: pages **p25-p29** written 07:12-07:32Z (decrypted: the phone-side conversation, Seeker device stamp)
- Desktop: kept appending its own **p25-p29+** for the same session through the afternoon
- Cloud now holds the **desktop's** content for those page keys (last flush wins); the phone's branch survives only in the phone's local tier

## Why it happens

`SessionStore.lastPageIndex()` picks the next page from what the opener can SEE at resume time; two live writers each think they own page N. Page keys are `{sessionId}__p{N}` with no writer fencing, so cloud uploads for the same key race and the loser's branch silently disappears from the cloud (still present in that device's local tier — nothing is unrecoverable today, but the cloud copy is a coin flip per page).

## Impact

- Silent divergence: each device sees a different history for the same session id; the cloud arbitrates per-page by upload order
- listMine's device label flips to whichever writer flushed the newest page last (this is what made #107 look like "phone sessions missing")

## Direction options (not decided)

1. **Single-writer lease**: a tiny `{sessionId}__writer` blob (device id + heartbeat ts); a second opener gets read-only or forks
2. **Fork-on-conflict**: second concurrent writer mints a new sessionId branching at the resume point (cheap, no coordination, matches the local-first model)
3. **Per-device page namespaces** + merge-on-read (heavier, but no data ever lost)

Related: #107 (how it was found), #100 (cross-surface QA).

RemilioNubilio:
I would take option 2. A lease (option 1) adds a coordination blob plus heartbeat expiry semantics, and a stale lease on a dead device blocks resume, which is the wrong failure mode for local-first. Option 3 changes the page key scheme and needs merge-on-read everywhere pages are decoded.

Option 2 is now mostly reuse: `SessionStore.fork()` landed in `packages/core/src/account/store.ts` (line 216) and already does the hard part, re-encoding every page under a fresh sessionId with correct per-page meta, with `upTo` to branch at the resume point. What is left is detection and wiring:

- at resume, remember the page index and message count the opener saw
[...]

PR #244, opened 2026-09-16
fix(feed): preserve comment drafts and recover failed reads

NubsCarson:
# Preserve mobile blog comment drafts until the post succeeds

Failed blog comments currently clear the draft immediately and can leave the composer stuck on Posting. An unrelated comments refresh also closes an inline reply while the user is typing. Handle the existing per-post blogCommentResult event instead: preserve inputs on failure, unlock retry, and clear only the successfully submitted composer. Also expose the host-supported fresh flag in the webview request type.

Four component regressions pass, including Korean input, late refreshes, repeated failures and an unrelated unsent draft. Two of these regressions fail on unchanged main. Production webview build passes. Full typecheck retains ten pre-existing unused-symbol failures; this patch adds none and removes the existing fresh request-type error. Local screenshots use simulated responses; no on-chain posts were submitted.

Addresses part of #208; does not close the issue. Session concurrency #113 and the separate pending-PR merge conflicts are not fixed by this patch.

### Local browser evidence

Simulated failure, with Korean draft retained:

Draft retained

390px mobile, reply retained after a late comments refresh:

Mobile reply

### Read recovery follow-up

This PR also discards superseded feed/comment responses, distinguishes failed reads from genuinely empty results, and adds Retry controls. A missing author-table body is never replaced with an unverified feed preview. A confirmed comment stays successful if only the comments refresh fails.

At `15d252e3`: 480 core tests, 7 webview tests, and production build passed. Ten baseline unused-symbol typecheck errors remain. The initial full-core run encountered an existing session-settings timing/cleanup race; both runs are preserved. No public posts or wallet transactions were submitted.

[Read recovery tests and limitations](https://github.com/NubsCarson/AgentNet/tree/15d252e3/docs/qa/blog-read-recovery)

Feed failure with Retry
Retry restored the feed
Missing author body


zo-sol:
Reviewed and cleared to merge on my side — the on-chain posting/signing path is untouched, no double-post risk, and the disclosed read-recovery changes look sound.

However, after the sibling PRs landed on main (#240 Windows npm engines, plus #241 and #242), this branch now conflicts and GitHub marks it `DIRTY`:
- `pnpm-lock.yaml` / `package.json` (overlaps with #240's dependency additions)
- the shared `store.tsx` / `protocol.ts` slices (overlaps with #241/#242)

Could you rebase onto the latest `main` and push? Once it's mergeable with green CI again I'll merge it. This is purely a mechanical conflict from the sibling PRs merging first — no review blockers.

